PR #9, opened 2016-04-04
Verified irma config. Added a missing envvar

sylvinite:
Made sure all envvar's look good from Scilife @ solna's side. Minor tweaks.


johanherman:
:+1:

PR #11, opened 2016-04-05
Added script for reference generation

sylvinite:
Also updated config for upcoming mail conf.

Code is probably not structured as it should be. Input is appreciated.


host_vars/127.0.0.1
```diff
@@ -50,6 +50,7 @@ charon_base_url: http://nestor1:4242
 # FIXME: hg19 was empty though. Needs to fetch from somewhere else.
```
johanherman: This FIXME I wrote when I was unsure if we would need hg19, but I'm now guessing we don't, right?

roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2
```diff
@@ -85,3 +85,5 @@ paths: # Hard code paths here if you are that kind of a person
 logging: 
     log_file: "{{ ngi_pipeline_logs }}/ngi_pipeline.log" 
 
+mail:
+    recipient: ngi_pipeline_operators@scilifelab.se
```
johanherman: Not strictly necessary, but if you think we want to change this value easier in the future, or have separate versions for Stockholm and Uppsala, then adding this as a variable from either `ngi_pipeline/defaults/` or `host_vars` might be suitable.

roles/piper/tasks/main.yml
```diff
@@ -78,5 +78,6 @@
   template: src="uppmax_global_config.xml.j2" dest="{{ ngi_piper_bin_dest }}/workflows/uppmax_global_config.xml"
   tags: ngi_piper 
 
-
+- name: deploy script (gen_GATK_ref.sh) for manual GATK reference generation
+  template: src="gen_GATK_ref.sh" dest="{{ ngi_piper_bin_dest }}/gen_GATK_ref.sh"
```
johanherman: Add a `tags: ngi_piper` here, so that the playbook can be run with the `--tags ngi_pipeline` argument, if an operator doesn't want to run all the roles in the playbook.

roles/piper/templates/gen_GATK_ref.sh
```diff
@@ -0,0 +1,13 @@
+module load bioinfo-tools
```
johanherman: A "good habit" is to name all templates with the `.j2` file extension (short for jinja2, the expression engine used for e.g. templates), and then when you're deploying them in your task you rename the file in the `dest` argument to be without the `j2` part.

johanherman:
:+1:

PR #12, opened 2016-04-05
Robbed funk_004 and applied to funk_006

sylvinite:
I'm a criminal, crim-i-naaaal.


roles/funk_006/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+ngi_resources: /lupus/ngi/ngi_resources/
+supervisord_log_dest: /home/funk_006/log/ 
+ngi_pipeline_web_port: 6666
```
johanherman: This port needs to be changed so it doesn't interfere with the Uppsala port.

PR #13, opened 2016-04-05
Initial support for multiple NGI site configs.

johanherman:
@sylvinite it seems to work fine on a `ansible-playbook deploy.yml --tags ngi_pipeline`. Next step would be to test a whole deploy, and then of course to see how it works inside the cluster. Do you have any comments on this so far?

### DIFF
--- a/host_vars/127.0.0.1
+++ b/host_vars/127.0.0.1
@@ -41,7 +41,7 @@ charon_api_token_upps: FOOBAR
 #Referring to dev instance for safety purposes
 charon_base_url: http://charon-dev.scilifelab.se/
 
-gatk_bundle_b37_origin = "/sw/data/uppnex/reference/biodata/GATK/ftp.broadinstitute.org/bundle/2.8/b37/"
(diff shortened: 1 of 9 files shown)


sylvinite:
I don't see anything that lets the user select what variables to use. I.e. dump to sthlm_conf -> source sthlm_conf. I'm probably missing something though.

johanherman:
Not completely sure what your question is. But there is e.g. one template called `irma_ngi_config.yaml.j2`. This file we want to deploy twice: one for Uppsala, and one for Stockholm. 

This is done on lines 48-68 in https://github.com/johanherman/pipeline-provision/pull/13/files#diff-741a9e2d5a8a65782ee0a5ddd06e73e8R48 with two `template` tasks. 

To get different values for Uppsala and for Stockholm both template tasks are preceeded by two `set_facts` tasks that set the generic variables in the template to site specific variables. 

These site specific variables are in turn fetched from `roles/ngi_pipeline/defaults/main.yml`. 

Does that answer your question?

sylvinite:
Yea, didn't see L56 & L67 respectively.

PR #15, opened 2016-04-06
Added external charon file support

sylvinite:
I probably goofed up a lot. But it easier to make this PR and ask rather than copypaste on slack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.retry
 .vault_password.txt
 pontus.larsson_medsci.uu.se.key
+charon_credentials.yml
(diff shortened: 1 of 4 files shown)


johanherman:
I'm not sure where you are putting `charon_credentials.yml`, but I suggest you put it under `ngi_pipeline/default/`, and inside then have three variables: 

```
charon_base_url: http://charon-dev.scilifelab.se/
charon_api_token_upps: FOOBAR
charon_api_token_sthlm: FOOBAR2
```

These values will then be picked up by the templates and tasks in the role, and put into the respective sourcme-files when deployed.

PR #16, opened 2016-04-06
Charon credentials

johanherman:
Requires a file `charon_credentials.yml` to be put under `host_vars/127.0.0.1/`, because it wasn't possible to have several var files under `roles/ngi_pipeline/defaults` (a `main.yml` already exists there). 

As a consequence `host_vars/127.0.0.1.yml` is also renamed to `host_vars/127.0.0.1/main.yml`, and the gatk key is now supposed to be placed under `playbooks/files` instead of `roles/piper/files` to make it a little bit more similar to the charon credentials.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.retry
 .vault_password.txt
 pontus.larsson_medsci.uu.se.key
+charon_credentials.yml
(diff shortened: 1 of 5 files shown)


sylvinite:
Great stuff!

PR #18, opened 2016-04-07
Sync script + logging 

johanherman:
Initial version of a sync script.

Logs in over SSH to irma1 with password and extra factor to enable a grace period on the cluster when the extra factor is not needed. After that the script can initate an rsync transfer by just suppling the SSH password again. Rsync will ignore `.swp` files and `.git` directories. The rsync part of the sync is at the moment logged. The script doesn't handle wrong supplied password/factor at the moment. 

Also a commit for logging all Ansible runs.

These logs are at the moment placed under `/lupus/ngi/irma3/log/{ansible,rsync}.log`. I envision that we move other local stuff there (`ansible-env`, `playbooks`, etc), to make it more clear what is local to irma3 and what is supposed to be synced to the cluster. When that is done we can also simplify the rsync filter somewhat. 

Also adding the `bashrc` file from irma3 to the `bootstrap` folder for safekeeping. Important that the `ngi-sw` group gets set as GID, and that the umask is appropriate when installing/deploying sw.


sylvinite:
Looks good, nothing to complain about.

PR #19, opened 2016-04-07
Initial taca commit

sylvinite:
Should get taca running. First try, so some things aren't what they should be.

### DIFF
--- a/roles/taca/tasks/main.yml
+++ b/roles/taca/tasks/main.yml
@@ -1,0 +1,27 @@
+- name: Fetch TACA-ngi-pipeline from GitHub
+  git: repo={{ taca_ngi_repo }}
+       dest={{ tca_ngi_dest }}
+       version={{ taca_ngi_version }}
+       force=yes
+  tags: taca
+
+- name: Possibly setup virtual environment and install dependencies 
+  shell: conda create -n NGI pip sqlalchemy 
+  when: not ngi_venv_exists.stdout 
+  tags: taca
+
+- name: Install TACA-ngi-pipeline requirements
+  shell: /lupus/ngi//sw/anaconda/envs/NGI/bin/pip install -r /lupus/ngi/sw/taca-ngi-pipeline/requirements.txt
+  tags: taca
+
+- name: Install TACA-ngi-pipeline
+  shell: cd {{ taca_ngi_dest }} && /lupus/ngi/sw/anaconda/envs/NGI/bin/pip install .
+  tags: taca
+
+- name: Deploy taca_delivery.yml config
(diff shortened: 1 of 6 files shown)


roles/taca/templates/taca_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
+log:
+    file: "{{ ngi_pipeline_logs }}/taca.log"
+
+deliver:
+    rootpath: "/proj/{{ ngi_pipeline_sthlm_delivery }}/nobackup/NGI/analysis_ready"
```
johanherman: From this I presume we need two versions of taca_delivery.yml deployed? One for Ua and one for Sthlm.

roles/taca/templates/taca_fastq_delivery.yml.j2
```diff
@@ -0,0 +1,21 @@
+log:
+    file: "{{ ngi_pipeline_logs}}/taca.log"
+
+deliver:
+    rootpath: "{{{{ ngi_pipeline_sthlm_delivery }}/nobackup/NGI/analysis_ready"
```
johanherman: Same thing here apparently.

roles/taca/templates/taca_fastq_delivery.yml.j2
```diff
@@ -0,0 +1,21 @@
...
+    analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
+    datapath: <ROOTPATH>/DATA/<PROJECTID>
+    stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
+    deliverypath: /proj/<UPPNEXID>/INBOX/<PROJECTID>
+    report_aggregate: ngi_reports ign_aggregate_report -n stockholm --pandoc_binary
```
johanherman: And if two versions are needed then "stockholm" should probably be a variable that can be set to "uppsala" in the other case.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+- name: Fetch TACA-ngi-pipeline from GitHub
+  git: repo={{ taca_ngi_repo }}
+       dest={{ tca_ngi_dest }}
+       version={{ taca_ngi_version }}
+       force=yes
+  tags: taca
+
+- name: Possibly setup virtual environment and install dependencies 
+  shell: conda create -n NGI pip sqlalchemy 
+  when: not ngi_venv_exists.stdout
```
johanherman: If you're going down this route then this will require that the ngi_pipeline role has been run before, otherwise this task will fail to find `ngi_venv_exists`.

sylvinite:
Closing this for now. Needs to be refactored due to other development. We'll also wait on it until after ngi_pipeline proper has been deployed.

PR #25, opened 2016-04-08
Setgid on all directories we create 

johanherman:
Have asked UPPMAX to remove setgid on root folder /lupus/ngi, so that we can have subdirs owned by both ngi-sw and ngi. This requires that roles that creates directories in some task manually enforce the setgid on the respective directory though.


sylvinite:
Looks good.

PR #26, opened 2016-04-08
Move db log

johanherman:
- Mention required setgid changes and don't keep backups of confs 
- Create log and db dir under project areas instead. Requires that the script `/lupus/ngi/resources/create_ngi_pipeline_dirs.sh` is run by a funk account with a specific project id as argument. 
- Use the delete flag when syncing.


sylvinite:
Yes! We can do it!

PR #27, opened 2016-04-08
Anaconda path

johanherman:
- correct swp filter in find command in sync.py
- use the ngi venv variable instead of using the hard coded anaconda path


sylvinite:
Looks guuud

PR #28, opened 2016-04-11
Update script to generate gatk indices and symlink in the reference data

johanherman:
Also update the README with info about anaconda and setgid.
And add an exception for the sync of the gatk_bundle


sylvinite:
Looks good, merging.

PR #29, opened 2016-04-11
Install supervisord into NGI venv and refactor func account role

johanherman:


sylvinite:
Looks good; I like the new layer of seperation.

PR #33, opened 2016-04-12
Tarzan web proxy

johanherman:
Tarzan -- a role for deploying the Kong web proxy (http://getkong.org). Also deploys dependency products, such as LuaJIT, Luarocks, OpenResty and Cassandra. 

The role successfully compiles and installs all sw on irma3, and sets up an initial version of a supervisord conf for handling e.g. Kong. After a sync to the cluster the kong binary can be launched, but fails due to wrongly configured Cassandra. 

Some things left to do: 
- Configure Cassandra properly 
- Add Cassandra to supervisord
- Investigate whether the Cassandra and Kong supervisord should be merged with one of the already existing supervisords. (Perhaps not? We're already using one for Uppsala atm, and one for Stockholm)
- Contemplate whether we should have a common tarzan config area under /lupus/conf/. (Atm these logs are written under Uppsala project dir, otherwise they will be overwritten.)

When these things finally are working one can continue and config up the backend services for the web proxy, and add things like SSL certificates, user tokens, etc.


sylvinite:
No complaints here.

PR #36, opened 2016-04-13
Confirm sync with a dry-run first.

johanherman:


sync.py
```diff
@@ -127,7 +127,26 @@ def yes_or_no(question):
...
+# First doing a dry-run to confirm sync. 
+print('Initiating a rsync dry-run')
+child = pexpect.spawn(dry_cmd)
+child.expect(exp_pass) 
+print('Sending dry-run password')
+child.sendline(password)
+child.interact()
+child.close()
+
+choice = yes_or_no("Do you want to do this sync for real? ")
```
sylvinite: The prompt is a bit confusing. I'd prefer something more along the lines of "Dry run finished. Do you wish to perform an actual sync with the same files?"

johanherman: Aight. Fixing.

PR #40, opened 2016-04-13
Taca stuff

sylvinite:
Initial commit. Pretty much to be desired. Feedback is appreciated.


roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+       version={{ taca_ngi_version }}
+       force=yes
+  tags: taca
+
+- name: Install TACA-ngi-pipeline requirements
+  shell: /lupus/ngi//sw/anaconda/envs/NGI/bin/pip install -r /lupus/ngi/sw/taca-ngi-pipeline/requirements.txt
+  tags: taca
+
+- name: Install TACA-ngi-pipeline
+  shell: cd {{ taca_ngi_dest }} && /lupus/ngi/sw/anaconda/envs/NGI/bin/pip install .
```
johanherman: Same thing here as on L9. 

You could also split out the module arguments like: 

```
shell: (...)pip install 
args: 
  chdir: {{ taca_ngi_dest }}
```

to make it a bit more readable and easy to add more arguments in the future.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+  tags: taca
+
+-name: Create TACA delivery configs directory
+  file: path="{{ ngi_resources }}/TACA" state=directory mode=g+s
+  tags: taca
+
+- name: Deploy sthlm_taca_delivery.yml config
```
johanherman: Instead of having two different copies of all the templates, where we in one copy use the Uppsala variables, and in the other copy use the Stockholm varibles, I suggest an other approach that we're using in the `ngi_pipeline` and `func_account` roles: 

- have a generic template, with a generic variable name 
- have site specific variables under the `defaults/main.yml` for Uppsala and Stockholm
- before the `template` task, use a `set_fact` task, that sets the generic variable in the template file to the site specific variable that you defined under `defaults` 

That way it gets a bit cleaner and organized -- with just one set of files, but different with variables passed into them at different times.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+- name: Fetch TACA-ngi-pipeline from GitHub
+  git: repo={{ taca_ngi_repo }}
+       dest={{ taca_ngi_dest }}
+       version={{ taca_ngi_version }}
+       force=yes
+  tags: taca
+
+- name: Install TACA-ngi-pipeline requirements
+  shell: /lupus/ngi//sw/anaconda/envs/NGI/bin/pip install -r /lupus/ngi/sw/taca-ngi-pipeline/requirements.txt
```
johanherman: Use `{{ ngi_venv }}` and `{{ taca_ngi_dest }}` or similar.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+       version={{ taca_ngi_version }}
+       force=yes
+  tags: taca
+
+- name: Install TACA-ngi-pipeline requirements
+  shell: "{{ ngi_pipeline_venv }}/bin/pip install" -r "{{ taca_ngi_dest }}/requirements.txt"
+  tags: taca
+
+- name: Install TACA-ngi-pipeline
+  shell: cd "{{ taca_ngi_dest }}" && "{{ ngi_pipeline_venv }}/bin/pip" install .
```
johanherman: I've never tried running a `shell` command with mixed quotes like that. I've always quoted the whole string, or nothing. Do you know if this will work?

sylvinite: No idea, I'll just quote the whole thing then.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  tags: taca
+
+-name: Create TACA delivery configs directory
+  file: path="{{ ngi_resources }}/TACA" state=directory mode=g+s
+  tags: taca
```
johanherman: `file` and `tags` shouldn't be indented. I guess it would make more sense to deploy the configs under `{{ ngi_pipeline_conf }}` as well, right?

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+
+- set_fact: 
+    site: "upps"
+  tags: taca
+
+- name: Deploy upps_taca_delivery.yml config
+  template: src="upps_taca_delivery.yml.j2" dest="{{ ngi_resources }}/TACA/{{ site }}_taca_delivery.yml"
```
johanherman: Some comment here about `ngi_resources` vs `ngi_pipeline_conf`. 

`upps_taca_delivery.yml.j2` should be `site_taca_delivery.yml.j2`.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  tags: taca
+
+- set_fact: 
+    site: "sthlm"
+  tags: taca
+
+- name: Deploy sthlm_taca_delivery.yml config
+  template: src="sthlm_taca_delivery.yml.j2" dest="{{ ngi_resources }}/TACA/{{ site }}_taca_delivery.yml"
+  tags: taca
+
```
johanherman: Same comment here for both these tasks about `ngi_resources` and the name of the src file.

roles/taca/templates/site_taca_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
+log:
+    file: "{{ ngi_pipeline_{{site}}_path }}/log/taca.log"
```
johanherman: I have never tried doing nested variable expansions like this. Do you know if this will work? 

Otherwise use a generic variable name in the template that you set to a site-specific value in the `set_fact` before the `template` task.

roles/taca/templates/site_taca_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
...
+
+deliver:
+    rootpath: "/proj/{{ ngi_pipeline_{{site}}_delivery }}/nobackup/NGI"
+    analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
+    datapath: <ROOTPATH>/DATA/<PROJECTID>
+    stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
+    deliverypath: /proj/<UPPNEXID>/INBOX/<PROJECTID>
+    report_aggregate: ngi_reports ign_aggregate_report -n stockholm --pandoc_binary
```
johanherman: Should `stockholm` be a variable?

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+- name: Deploy sthlm_fastq_taca_delivery.yml config
+  template: src="sthlm_fastq_taca_delivery.yml" dest="{{ ngi_resources }}/TACA/sthlm_taca_fastq_delivery.yml"
```
johanherman: You're missing the suffix `.j2` in the src file.

roles/taca/templates/sthlm_fastq_taca_delivery.yml.j2
```diff
@@ -0,0 +1,21 @@
+log:
```
johanherman: Is this file only needed for Stockholm, or will there be one for Uppsala as well?

roles/taca/templates/sthlm_fastq_taca_delivery.yml.j2
```diff
@@ -0,0 +1,21 @@
...
+    file: "{{ ngi_pipeline_sthlm_path }}/log/taca.log"
+
+deliver:
+    rootpath: "{{{{ ngi_pipeline_sthlm_delivery }}/nobackup/NGI"
```
johanherman: Too many `{{` here.

roles/taca/tasks/main.yml
sylvinite: No. Since {{ ngi_pipeline_conf }} contains more or less static configuration whiles wheras taca uses its "conf" as an argument every time it is called, it'd be wrong to put it under conf. I contemplated making a seperate directory called taca_configs, but I felt it was a bit excessive.

roles/taca/templates/sthlm_fastq_taca_delivery.yml.j2
sylvinite: upps_fastq_taca_delivery has never existed will not exist in the forseeable future.

johanherman: Gotcha. :+1:

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- set_fact: 
+    site: "upps"
+    site_full: "uppsala"
+    ngi_pipeline_site_path: {{ ngi_pipeline_upps_path }}
+    ngi_pipeline_site_delivery: {{ ngi_pipeline_upps_path }}
+  tags: taca
+
+- name: Deploy site_taca_delivery.yml config
+  template: src="{{ site }}_taca_delivery.yml.j2" dest="{{ ngi_pipeline_conf }}/TACA/{{ site }}_taca_delivery.yml"
```
johanherman: `src` should be `site_taca_delivery.yml.j2`, because we do not want to expand `site` in this case.

roles/taca/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- set_fact: 
+    site: "sthlm"
+    site_full: "stockholm"
+    ngi_pipeline_site_path: {{ ngi_pipeline_sthlm_path }}
+    ngi_pipeline_site_delivery: {{ ngi_pipeline_sthlm_delivery }}
+  tags: taca
+
+- name: Deploy site_taca_delivery.yml config
+  template: src="{{ site }}_taca_delivery.yml.j2" dest="{{ ngi_pipeline_conf }}/TACA/{{ site }}_taca_delivery.yml"
```
johanherman: Same thing here. `site` should not be expanded in `src`.

johanherman:
Great work! :+1: Merging!

PR #41, opened 2016-04-13
Cassandra conf

johanherman:
Installs an other Kong requirement: serf. 

Initial config of Cassandra to at least make it boot up sensible. 

Have verified by starting it manually in foreground with `cassandra -f` on the cluster, and a manual start of Kong with `kong -c /lupus/ngi/conf/webproxy.yml` at least communicates with it and seems to setup the relevant data storage entities. 

The current show stopper is that Kong afterwards aborts with a: 

```
[ERR] nginx: [emerg] at least OpenSSL 1.0.2e required but found OpenSSL 1.0.1e-fips 11 Feb 2013
[ERR] Could not start Kong
```

### DIFF
--- a/roles/tarzan/README.md
+++ b/roles/tarzan/README.md
@@ -2,4 +2,9 @@
 
 Ansible role for deploying the Kong (https://getkong.org/) API and microservice management layer. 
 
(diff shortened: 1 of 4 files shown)


sylvinite:
Nothing to complain about.

PR #42, opened 2016-04-14
Newer version of openssl for openresty

johanherman:
- Compile openresty with newer version of openssl than the one provided from the system. 
- Set correct `LUA_PATH` and `LUA_CPATH` env vars for funk_004 user.

These two changes gets Kong working. After running

- `cassandra -f `
- `kong -c /lupus/ngi/conf/webproxy.yml`

a `curl localhost:8001` will return a JSON payload from the Kong server.

### DIFF
--- a/roles/tarzan/defaults/main.yml
+++ b/roles/tarzan/defaults/main.yml
@@ -33,6 +33,12 @@ openresty_file: openresty-1.9.7.3.tar.gz
 openresty_version: 1.9.7.3
 openresty_dest: "{{ tarzan_dest }}/openresty/"
 
+# openresty depends on a newer version of openssl
+openssl_url: https://www.openssl.org/source/openssl-1.0.2f.tar.gz
+openssl_file: openssl-1.0.2f.tar.gz
+openssl_version: 1.0.2f
+openssl_dest: "{{ tarzan_dest }}/openssl/"
+
 # kong depends on serf for service/cluster discovery 
 serf_url: https://releases.hashicorp.com/serf/0.7.0/serf_0.7.0_linux_amd64.zip
 serf_file: serf_0.7.0_linux_amd64.zip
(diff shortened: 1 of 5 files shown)


roles/tarzan/defaults/main.yml
```diff
@@ -33,6 +33,12 @@ openresty_file: openresty-1.9.7.3.tar.gz
 openresty_version: 1.9.7.3
 openresty_dest: "{{ tarzan_dest }}/openresty/"
 
+# openresty depends on a newer version of openssl
+openssl_url: https://www.openssl.org/source/openssl-1.0.2f.tar.gz
```
sylvinite: I suggest you use /source/old/ so the URL is less likely to break.

roles/tarzan/tasks/main.yml
```diff
@@ -16,9 +16,20 @@
               backup=no
   tags: tarzan
 
+- name: add the lua_path to funk_004's environment via the sourceme_upps script
+  lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_upps_script }}
+              line='export LUA_PATH={{ lua_path }}'
+              backup=no
```
sylvinite: If we really want to go this route, wouldn't it make more sense for ngi_pipeline and charon to add their respective paths in the same way?

roles/tarzan/defaults/main.yml
johanherman: Ah. Good suggestion. Will add that.

roles/tarzan/tasks/main.yml
johanherman: In what way do you mean? We do add env variables for ngi_pipeline and charon "In the same way" as in adding common variables to the common sourceme script, and site specific variables to the site specific sourcme script.

sylvinite: If I understand you correctly this writes the entire export line for the lua_path.
For charon & ngi_pipeline that export line always exists, the deployment merely adds the relevant paths.

johanherman: Ah! OK. Yeah, that is true. I understand what you mean. 

It would be a bit more "consistent" if we created these scripts by only using e.g. the 'lineinfile` module, but that would make it a bit more messy if we want to have static content in the files as well. 

In this case I guess it is mostly the comment lines in the sourcme scripts. We would then have to create a `lineinfile` task for each of these comments, and also having two tasks for the site scripts that does the `source` of the common script. 

One of the drawbacks of having a mix of `lineinfile` and templates is that the `ngi_pipeline` role needs to run before the `tarzan` role, as otherwise `tarzan` would create the Uppsala script, and then if someone runs the `ngi_pipeline` role afterwards then the script will be overwritten by the template. 

Same thing applies for the `piper` role, because that adds the PATHs via `lineinfile` as well: if `piper` is run without `ngi_pipeline` having been run beforehand, and then someone decides to run `ngi_pipeline` afterwards, then the sourcme script will be overwritten. 

One can add a `meta/main.yml` for the `tarzan` and `piper` roles though, that says that `ngi_pipeline` is a dependency. That will cause the `ngi_pipeline` to always run before running the role in question. We should perhaps add this.

sylvinite: Yes. Either we consider ngi-pipeline some kind of boss (like you said with the dependency), or we break out the sourceme into a seperate function. Either one works.

I'll merge this for now and we can add that whenever there's time/interest.

PR #43, opened 2016-04-14
Deploy checksum service

johandahlberg:
This is a preliminary playbook and role used to install the checksum service. Still needs to have supervisord, etc setup. But I'll leave that to someone else. :wink:


johanherman:
Merging this for now.

PR #45, opened 2016-04-15
Tarzan tuning

johanherman:
- Make serf and Kong admin interface only listen to localhost.
- Change kong https port to 4444 and non-https to 8888
- Need escaping of lua env vars to not get bash errors 
- Need to add some system lua paths to the lua envs for module system lmod to function, as it is also using lua. Might be a future error source.
- Spawn 4 nginx worker processes instead of the default "auto", which on irma1 will spawn 16 processes (number of CPU cores)


sylvinite:
Nothing too crazy, merging.

PR #51, opened 2016-04-18
Taca playbook

sylvinite:
Hopefully works as intended. Probably a few test deliveries need to be done; but ngi_reports need to be implemented to get a proper picture.


johanherman:
Looks good from what I can tell. :+1: Will merge. 

A general comment though so I don't forget: 

As you might remember I changed from using the Ansible module's `pip` module to `shell` and there running the pip command manually, because we had some problems with anaconda and then the pip module after that. 

We should probably "someday" try to see if the pip module works now when the anaconda dir is not setgid any longer. I'm not sure I tried that. 

The problem with running too many shell commands in the playbooks is that the shell command will run every time you run the play book (if you don't use the `creates=` argument to the `shell` task). This can cause problems with e.g. files getting updated even if they haven't changed etc. 

An other general comment: 

When you're developing the roles you might want to start running the sync (but abort before pushing anything). That way you will e.g. see if you accidentally have created some files with wrong permissions. (Tarzan have a bunch of files with wrong permissions at the moment for example.)

PR #54, opened 2016-04-20
Cron fix

sylvinite:
Small stuff


johanherman:
Need to wait a whole hour for the next crontab to take effect?! All right then.

PR #56, opened 2016-04-21
Piper bug? Symlink needed for setupfilecreator

johanherman:


sylvinite:
Sicken snusk!

Looks good, merging.

PR #58, opened 2016-04-21
Overriding default setupfilecreator path

sylvinite:


johanherman:
Sweet.

You could actually also remove the setupfilecreator stuff from the task that adds it to `$PATH` as well.

johanherman:
:+1:

PR #60, opened 2016-04-22
Don't try to update the timestamps for directories, 

johanherman:
as that will cause a lot of errors and warnings when an operator tries to update a directory created by an other user.


sylvinite:
Tested and no more complaints, horray!

PR #63, opened 2016-04-25
Tarzan supervisor

johanherman:
Set the file permission correctly for everything installed under `sw/tarzan/`. A ugly hack is necessary on L68-70 in `tasks/main.yml` that sets the permissions for group and others manually, because it seems some of the installation/compilation steps doesn't respect the umask properly set (?!). This is only done on the files owned by the current user, to limit the amount of errors. So it is expected that each user that runs this role actually runs the complete role so that the permissions gets updated for all new files added. 

Installs the `serf` software that helps kong with cluster discovery. 

Setups proper logging directory for cassandra. 

Launches cassandra via Uppsala's supervisord instance, but it doesn't seem to work easily for kong due to how it spawns sub-processes (nginx + serf) + daemonizes. Instead we're launching kong manually via Uppsala's crontab.

### DIFF
--- a/roles/tarzan/defaults/main.yml
+++ b/roles/tarzan/defaults/main.yml
@@ -51,6 +51,7 @@ tarzan_env_path: "{{ luajit_dest }}/{{ luajit_version }}/bin:{{ luarocks_dest }}
 
 tarzan_log_dest: "{{ ngi_pipeline_upps_path }}/log/tarzan/"
 tarzan_db_dest: "{{ ngi_pipeline_upps_path }}/db/tarzan/"
+cassandra_log_dest: "{{ tarzan_log_dest }}/cassandra"
 cassandra_db_dest: "{{ tarzan_db_dest }}/cassandra" 
 
 # To get Tarzan to work lua_path and lua_cpath were generated from a working luarocks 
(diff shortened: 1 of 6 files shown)


roles/tarzan/defaults/main.yml
```diff
@@ -51,6 +51,7 @@ tarzan_env_path: "{{ luajit_dest }}/{{ luajit_version }}/bin:{{ luarocks_dest }}
 
 tarzan_log_dest: "{{ ngi_pipeline_upps_path }}/log/tarzan/"
 tarzan_db_dest: "{{ ngi_pipeline_upps_path }}/db/tarzan/"
+cassandra_log_dest: "{{ tarzan_log_dest }}/cassandra"
```
sylvinite: Might want to consider moving out cassandra's log & db from /tarzan/. Depends on how hard the dependency is I suppose. Not a critique per se, just a consideration.

sylvinite:
Looks really good. If you want to keep the cassandra logs & db where you put them I'll just merge it.

johanherman:
There's no dependency per se where I put the cassandra db and logs. I have just put them under the tarzan sub folder at the moment to make it more clear what it concerns. So I think you can go ahead and merge.

PR #66, opened 2016-04-26
Ngi reports initial commit

sylvinite:
Works to install. Still have to debug a few runtime issues.


README.md
```diff
@@ -33,6 +33,8 @@ The following files need to be present on irma3 in order to successfully deploy
 
 - A `charon_credentials.yml` file placed under `/lupus/ngi/irma3/host_vars/127.0.0.1/` listing the variables `charon_base_url`, `charon_api_token_upps` and `charon_api_token_sthlm`
 
+- A valid statusdb.yml access file placed under `/lupus/ngi/irma3/deploy/files`.
```
johanherman: You might want to mention what fields this file should include.

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -0,0 +1,76 @@
...
+  shell: "{{ ngi_pipeline_venv }}/bin/pip install couchdb"
+  tags: ngi_reports
+
+- name: Copy statusDB credentials
+  copy: src="files/{{ statusdb_creds }}" dest="{{ ngi_pipeline_conf }}/{{ statusdb_creds }}"
+  tags: ngi_reports
+
+- name: Add statusDB envvar to sourceme 
+  lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_sthlm_script }}
+              line='export STATUS_DB_CONFIG={{ ngi_pipeline_conf }}/statusdb.yaml'
```
johanherman: Should be `statusdb.yml`. Perhaps you can use your environment variable `statusdb_creds` instead.

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -0,0 +1,76 @@
...
+              backup=no
+  tags: ngi_reports
+
+###END STOCKHOLM SPECIFIX
+
+#NOTE: WORKAROUND MAY NOT BE REQUIRED, BUT LIKELY SO.
+- name: Copy Pandoc fix
+  copy: src=make_report.sh dest="{{ ngi_resources }}/make_report.sh"
```
johanherman: Is `make_report.sh` launched manually or via some other script/program? If the latter then you might have to make sure that the script is executable: either make sure you have done a `chmod +x make_report.sh` in the repo-checked in version of the file, or just use a `mode` flag to the `copy` to make sure that it is executable when installed. (I presume a manual launch can always be done by first typing `bash ..`...)

roles/ngi_reports/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
+---
+#- name: Check if dev_IS virtual env exists
```
johanherman: Might these uncommented tasks be needed in the future or should they be removed now?

roles/ngi_reports/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
...
+  args:
+    chdir: "{{ ngi_reports_dest }}/data/pandoc_templates/pandocfilters/"
+  tags: ngi_reports
+
+- set_fact:
+    site: "uppsala"
+  tags: ngi_reports
+
+- name: Create Uppsala ngi_reports conf file
+  template: src="ngi_reports.conf" dest="{{ ngi_pipeline_conf }}/ngi_reports.conf"
```
johanherman: You could rename the template on the source side to `ngi_reports.conf.j2` to make it clearer that it is a jinja2 template.

johanherman:
Had some comments, but otherwise it looks great. :+1:

sylvinite:
Fixed what you mentioned and also added some pretty minor necessities to run proper.

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -57,20 +57,20 @@
...
               backup=no
   tags: ngi_reports
 
-###END STOCKHOLM SPECIFIX
+###END STOCKHOLM SPECIFIC
 
 #NOTE: WORKAROUND MAY NOT BE REQUIRED, BUT LIKELY SO.
 - name: Copy Pandoc fix
-  copy: src=make_report.sh dest="{{ ngi_resources }}/make_report.sh"
+  copy: src=make_report.sh dest="{{ ngi_resources }}/make_report.sh" mode=0770
```
johanherman: Is `0770` really correct? That will set `u=rwx,g=rwx,o=`, i.e. no one outside of the `ngi-sw` group will be able to read/execute. I'm guessing you want `0775`. Or just use a `mode=a+x` if you know that the file has correct read/write permissions in the repo.

taca-inst.yml
```diff
@@ -4,5 +4,6 @@
   hosts: 127.0.0.1
   connection: local
   roles: 
-   - taca
+   #-ngi_pipeline
+   #- taca
```
johanherman: How are you planning that the installation of `taca` and `ngi_reports` is supposed to happen? Are they going to be in `install.yml`, or should we install them separately via this `taca-inst.yml`? Just wondering because you outcommented these two lines now.

PR #67, opened 2016-04-27
Cassandra 2.2.5 doesn't exist, updating to 2.2.6

sylvinite:
Updated cassandra version since 2.2.5 doesn't exist at the source.
If you want to pull 2.2.5 from another source instead, feel free to comment.

### DIFF
--- a/roles/tarzan/defaults/main.yml
+++ b/roles/tarzan/defaults/main.yml
@@ -22,9 +22,9 @@ luarocks_version: 2.3.0
 luarocks_dest: "{{ tarzan_dest }}/luarocks/"
 
 # kong depends on cassandra as backend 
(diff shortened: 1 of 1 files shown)


johanherman:
Ugh. I hate it when projects *remove* links! 

Try https://github.com/apache/cassandra/archive/cassandra-2.2.5.zip instead then. The Cassandra db already contains information. I rather not upgrade at the moment, as I'm not sure how it will go.

sylvinite:
No problem. Figured moving to 2.2.6 would probably cause more problems than it'd solve. Reworking it in a minute.

johanherman:
My bad. That was a link to source code. Take this URL instead: http://archive.apache.org/dist/cassandra/2.2.5/apache-cassandra-2.2.5-bin.tar.gz

sylvinite:
I'll just close this for now and let you work it out yourself. That's probably a lot easier than just relaying info to me.

PR #71, opened 2016-04-28
Arteria siswrap ws

johanherman:
Installs the arteria-siswrap web service that listen on port 10430. Siswrap acts as a wrapper around SNP&SEQ's Sisyphus software suit that we use e.g. for report generation. Depends on a bunch of Perl stuff so to make it easier I've added a required bootstrap step that installs the `cpanm` binary on irma3 under the `ansible-env/bin` directory as that's already on the `$PATH`, so it can be picked up and used by Ansible's cpanm module.


roles/arteria-siswrap-ws/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+# TODO: Perl that launches needs to have the correct PERL5LIB setup. 
+- name: modify uppsala's supervisord conf to start arteria-siswrap-ws 
+  ini_file: 
+    dest={{ ngi_pipeline_conf }}/supervisord_upps.conf
+    section=program:arteria-siswrap-ws
+    option=command
+    value="{{ arteria_siswrap_env_root }}/bin/siswrap-ws --configroot={{ arteria_siswrap_config_root }} --port={{ arteria_siswrap_port }}"
+    backup=no
+ 
+- name: modify uppsala's supervisord conf to start arteria-siswrap-ws
```
sylvinite: Since they're separate tasks I'd prefer if they had some difference in their names.

roles/arteria-siswrap-ws/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  ini_file: 
+    dest={{ ngi_pipeline_conf }}/supervisord_upps.conf
+    section=program:arteria-siswrap-ws
+    option=autorestart
+    value=true
+    backup=no
+
+- name: add PERL5LIB to uppsala's sourcme script
```
sylvinite: sourceme over sourcme

roles/arteria-siswrap-ws/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+arteria_install_path: "{{ root_path }}/sw/arteria"
+arteria_siswrap_env_root: "{{ arteria_install_path }}/siswrap_venv"
+arteria_siswrap_config_root: "{{ ngi_pipeline_conf }}/arteria/siswrap"
+arteria_siswrap_app_config: "{{ arteria_siswrap_config_root }}/app.config"
+arteria_siswrap_logger_config: "{{ arteria_siswrap_config_root }}/logger.config"
+
+arteria_siswrap_port: 10430
+arteria_siswrap_mail_from: johan.hermansson@medsci.uu.se
```
sylvinite: Kind of odd that the siswrap is mailing from your personal account rather than a more dedicated entry.

sylvinite:
Looks good. Only consistent thing I noticed is that lines occasionally lack that use variables lack " " around them. Made some nitpicky pointers as well.

If you're ready to merge just say so.

roles/arteria-siswrap-ws/defaults/main.yml
johanherman: Yeah.. we do not have a functional account for mailing our stuff, so this is what we've been doing for now.

johanherman:
Fixed your small comments. You can merge.

PR #74, opened 2016-05-10
NGI reports related updates

sylvinite:
Minor changes related to:
Readme
NGI Pipeline - Broke out Envvar name
NGI Reports - Added necessary log file support, some bugfixes

### DIFF
--- a/roles/ngi_pipeline/tasks/main.yml
+++ b/roles/ngi_pipeline/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 
-- name: fetch ngi_pipline from github 
+- name: Fetch ngi_pipeline from github 
   git: repo={{ ngi_pipeline_repo }} 
        dest={{ ngi_pipeline_dest }}
        version={{ ngi_pipeline_version }}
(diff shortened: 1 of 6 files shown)


roles/ngi_pipeline/tasks/main.yml
```diff
@@ -1,22 +1,22 @@
...
        force=yes
   tags: ngi_pipeline 
 
 
-- name: check if ngi virtual env exists 
-  shell: conda env list | awk '{print $1}' | egrep -e "^NGI$"
-  register: ngi_venv_exists
+- name: "Check if {{ NGI_venv_name }} virtual env exists" 
+  shell: "conda env list | awk '{print $1}' | egrep -e ^{{ NGI_venv_name }}$"
+  register: "{{ NGI_venv_name }}_venv_exists"
```
johanherman: Does this work? I.e have you verified that the `{{ NGI_venv_name }}` will be expanded in `register`? Because it didn't use to work in `name`, at least for older versions of Ansible.

roles/ngi_reports/templates/ngi_reports_sthlm.conf.j2
```diff
@@ -0,0 +1,32 @@
...
+[quality_threshold_hiseqx]
+<150 : 80
+>150 : 75
+[quality_threshold_miseq]
+<100 : 80
+100 : 75
+150 : 70
+>250 : 60
+[log]
+log_dir : {{ ngi_pipeline_log_sthlm }}
```
johanherman: These vars are defined like 

```
ngi_pipeline_log_sthlm: "{{ ngi_pipeline_sthlm_path }}/log/ngi_pipeline.log"
```

Does that really match how `log_dir` in ngi_reports should be set?

roles/ngi_pipeline/tasks/main.yml
sylvinite: The lines you highlighted worked as expected.
If there's a known problem that register doesn't expand variables then I can just change it to a more general variable name. Is that the case though?

roles/ngi_reports/templates/ngi_reports_sthlm.conf.j2
sylvinite: Yea. It's bloody awful.

roles/ngi_pipeline/tasks/main.yml
johanherman: No, I don't know, that's why I asked. I've never tried doing it in this way before. If you have tested that it works OK with different values of `NGI_venv_name` then all OK from my side.

sylvinite: Ok sweet. 
I haven't seen any issue with it. There might, but it hasn't hit me yet at least.

PR #75, opened 2016-05-10
Taca updates for delivering

sylvinite:
Apparently the normal config files can't be used for "fake" deliveries (deliveries within irma). So adding all this stuff for the very small difference.


roles/taca/templates/site_fake_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
...
+        -
+            - <REPORTPATH>/delivery/reports/<PROJECTID>_aggregate_report.csv
+            - <STAGINGPATH>/00-Reports
+            - no_digest: True
+        -
+            - <REPORTPATH>/delivery/reports/<SAMPLEID>_ign_sample_report.html
+            - <STAGINGPATH>/<SAMPLEID>/00-Reports
+            - no_digest_cache: True
+        - 
+            - <ANALYSISPATH>/piper_ngi/07_variant_calls/<SAMPLEID>*.snpEff.summary.*
```
b97pla: unless there is a good reason not to, I think you should include a `.` after `<SAMPLEID>` to make sure that the complete sample id matches (e.g. `Sample1.suffix` should match but not `Sample10.suffix`)

roles/taca/templates/site_fake_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
...
+    operator: "{{ recipient_mail }}"
+    hash_algorithm: md5
+    files_to_deliver:
+        -   
+            - <ROOTPATH>/DELIVERY.README.txt
+            - <STAGINGPATH>
+        -
+            - <LOGPATH>/version_report.txt
+            - <STAGINGPATH>/00-Reports
+            - required: True
```
b97pla: In Uppsala, we have some more reports that are delivered, perhaps these could be included with a conditional?

> 
        -
            - <ANALYSISPATH>/qc_sisyphus/reports
            - <STAGINGPATH>/00-Reports/SequenceQC/Sisyphus

b97pla:
@sylvinite I had some comments above but they also apply to the "hot" taca configs, so could you modify them as well?

roles/taca/templates/site_fake_delivery.yml.j2
```diff
@@ -0,0 +1,85 @@
+log:
+    file: "{{ ngi_pipeline_site_path }}/log/taca.log"
+
+deliver:
+    rootpath: "/proj/{{ ngi_pipeline_site_delivery }}/nobackup/NGI"
+    analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
+    datapath: <ROOTPATH>/DATA/<PROJECTID>
+    stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
+    deliverypath: /proj/<UPPNEXID>/INBOX/<PROJECTID>
```
b97pla: On Irma4, this delivery path should currently be `/pica/sw/links/proj/<UPPNEXID>/INBOX/<PROJECTID>`

roles/taca/tasks/main.yml
```diff
@@ -26,17 +26,28 @@
...
+    proj_path: "/proj/"
+    extra_reports:"     -\n        - <ANALYSISPATH>/qc_sisyphus/reports\n        - <STAGINGPATH>/00-Reports/SequenceQC/Sisyphus"
   tags: taca
 
 - name: Deploy upps_taca_delivery.yml config
   template: src="site_taca_delivery.yml.j2" dest="{{ ngi_pipeline_conf }}/TACA/{{ site }}_taca_delivery.yml"
   tags: taca
 
+- set_fact:
+    proj_path: "/pica/sw/links/proj/"
```
b97pla: should there be a `tags: taca` statement here as well (and after line 62 below)?

roles/taca/tasks/main.yml
```diff
@@ -46,3 +57,10 @@
 - name: Deploy sthlm_fastq_taca_delivery.yml config
   template: src="sthlm_fastq_taca_delivery.yml.j2" dest="{{ ngi_pipeline_conf }}/TACA/sthlm_taca_fastq_delivery.yml"
   tags: taca
+
+- set_fact:
+    proj_path: "/pica/sw/links/proj/"
+
+- name: Deploy sthlm_fake_delivery.yml config
```
b97pla: I would have assumed that the "fake delivery" would be the within-Irma delivery with `proj_path: /proj` and the regular config pointing to `/pica/...` but it seems to be the other way around, is that the intention?

roles/taca/templates/site_taca_delivery.yml.j2
```diff
@@ -6,7 +6,7 @@ deliver:
     analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
     datapath: <ROOTPATH>/DATA/<PROJECTID>
     stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
-    deliverypath: /pica/sw/links/proj/<UPPNEXID>/INBOX/<PROJECTID>
+    deliverypath: "{{ proj_path }}}<UPPNEXID>/INBOX/<PROJECTID>"
```
b97pla: there's one `}` too many there if I'm not mistaken. since I'm nitpicking, I'd suggest to put the `/` here, rather than in the `proj_path` variable (and it won't break anything if it's in both positions)

b97pla:
:+1: looks great!

Issue #77, opened 2016-05-11
Set english locale for interactive sessions


b97pla:
When running an interactive session as funk_004, the locale is set to swedish. This causes e.g. pipeline output to have ',' as decimal separator etc. which messes up the parsing of the results. The `~/.bashrc` should be set up to always load english locale by default.

PR #78, opened 2016-05-11
Force English locale to get correct decimal separator in pipeline output

johanherman:
Solves https://github.com/NationalGenomicsInfrastructure/irma-provision/issues/77


sylvinite:
Looks alright

PR #84, opened 2016-05-16
Solved multiqc timeout error

sylvinite:
MultiQC was superslow since it tried to connect to the net. Added config file to go around it.

### DIFF
--- a/roles/multiqc/tasks/main.yml
+++ b/roles/multiqc/tasks/main.yml
@@ -9,8 +9,8 @@
   tags: multiqc
 
 - name: Force config upon multiqc command (sourceme script)
-  lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_script }}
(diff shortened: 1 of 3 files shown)


roles/multiqc/tasks/main.yml
```diff
@@ -3,4 +3,16 @@
...
   shell: "{{ ngi_pipeline_venv }}/bin/pip install multiqc --upgrade"
   tags: multiqc
+
+- name: Deploy multiqc config
+  copy: src="multiqc_config.yml" dest="{{ ngi_pipeline_conf }}"
+  tags: multiqc
+
+- name: Force config upon multiqc command (sourceme script)
+  lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_script }}
+              line='alias multiqc=multiqc -c {{ ngi_pipeline_conf }}/multiqc_config.yml'
```
johanherman: You should probably make this into `line="alias multiqc='multiqc -c {{ ngi_pipeline_conf }}/multiqc_config.yml'"` or something.

PR #85, opened 2016-05-17
ngi reports config fix

sylvinite:
NGI reports is sure being picky today.

### DIFF
--- a/roles/ngi_reports/templates/ngi_reports_site.conf.j2
+++ b/roles/ngi_reports/templates/ngi_reports_site.conf.j2
@@ -34,7 +34,7 @@ dm3 = Drosophila_melanogaster
 >250 : 60
 [log]
 {% if "sthlm" in site %}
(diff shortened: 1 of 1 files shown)


roles/ngi_reports/templates/ngi_reports_site.conf.j2
```diff
@@ -34,7 +34,7 @@ dm3 = Drosophila_melanogaster
 >250 : 60
 [log]
 {% if "sthlm" in site %}
-"log_dir : {{ ngi_pipeline_log_sthlm }}"
+log_dir : {{ ngi_pipeline_log_sthlm }}
 {% elif "upps" in site %}
 "log_dir : {{ ngi_pipeline_log_upps }}"
```
johanherman: Same fix should be applied on L39 then.

sylvinite: I wondered where that went!

PR #86, opened 2016-05-17
Support production and staging versions of Arteria services

johanherman:
Sets up production and staging specific paths and values in the form of `/lupus/ngi/conf/arteria/{checksum,siswrap}/{production,staging}` and `/lupus/ngi/sw/arteria/{checksum,siswrap}_{src,venv}/{production,staging}`, and updates Uppsala's `supervisord.conf` to listen on different ports depending on environment. 

What version to install is controlled via the vars `arteria_checksum_environment` and `arteria_checksum_version` (and corresponding for `siswrap`). Can be overridden on the command line when the operator wants to install a new version into staging e.g.

### DIFF
--- a/install_arteria.yml
+++ b/install_arteria.yml
@@ -11,6 +11,8 @@
   vars: 
     arteria_checksum_version: v1.0.3
     arteria_checksum_environment: production
+    arteria_siswrap_version: v1.0.2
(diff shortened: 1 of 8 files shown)


sylvinite:
Looks great!

PR #87, opened 2016-05-18
Support for MultiQC NGI module

sylvinite:
Added support for MultiQC NGI module.
Trimmed the code in some places.


roles/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+multiqc_ngi_repo: https://github.com/ewels/MultiQC_NGI.git
```
johanherman: I'm guessing you want to have this in `roles/multiqc/defaults/main.yml` :)

roles/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
...
+multiqc_ngi_dest: "{{ root_path }}/sw/multiqc_ngi"
```
johanherman: If the `multiqc_ngi` dir only exists as a placeholder for your checkout, then perhaps you could rename this to `multiqc_ngi_src` or something similar, to make it more clear that no sw is actually there. Perhaps we should even start using a `/lupus/ngi/src/` instead in the future.

roles/multiqc/tasks/main.yml
```diff
@@ -4,15 +4,29 @@
   shell: "{{ ngi_pipeline_venv }}/bin/pip install multiqc --upgrade"
```
johanherman: Should multiqc from pypi be installed at the same time as multiqc from the github repo?

roles/defaults/main.yml
sylvinite: Woops

roles/multiqc/tasks/main.yml
sylvinite: Yes. "MultiQC is from github" is actually a multiqc plugin.

PR #89, opened 2016-05-19
NGI Reports texlive installation

sylvinite:
TexLive wasn't properly installed in the past. This was ignored as NGI Reports had ten other things not making it functional. This update makes sure TexLive is actually installed and used.

Feel free to verify this a bit extra. The installation takes forever, so I've only quickly verified it once.

### DIFF
--- a/roles/ngi_reports/tasks/dependencies.yml
+++ b/roles/ngi_reports/tasks/dependencies.yml
@@ -2,18 +2,14 @@
 ##Plausible TODOs:
 # Install Helvetica Neue and Consolas from ngi_reports/data/pandoc_templates/assets/fonts/
 
-- stat: path="{{ texLive_dest }}/tlpkg/texlive.profile"
-  register: texlive_installed
-
 - name: Create TeXLive destination
   file: path={{ texLive_dest }} state=directory mode=g+rwxs
   tags: ngi_reports
 
-- name: Download TeXLive (Skipped if installed)
+- name: Download TeXLive
   get_url:
     url: "{{ texLive_url }}"
     dest: "{{ texLive_dest }}/src/{{ texLive_file }}"
-    when: not texlive_installed
     mode: g+rw
   tags: ngi_reports
 
@@ -29,6 +25,9 @@
 - name: Copy TexLive profile
   template: src="texlive.profile.j2" dest="{{ ngi_pipeline_conf }}/texlive.profile" mode=0775
   tags: ngi_reports 
+
+- stat: path="{{ texLive_dest }}/tlpkg/texlive.profile"
+  register: texlive_installed
 
 - name: Install TeXLive (Skipped if installed)
   shell: "./install-tl --profile={{ ngi_pipeline_conf }}/texlive.profile"
--- a/roles/ngi_reports/tasks/dependencies.yml
+++ b/roles/ngi_reports/tasks/dependencies.yml
@@ -2,21 +2,19 @@
 ##Plausible TODOs:
 # Install Helvetica Neue and Consolas from ngi_reports/data/pandoc_templates/assets/fonts/
 
-- name: Check for existing TeXLive installation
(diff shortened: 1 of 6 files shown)


roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -2,29 +2,44 @@
...
   file: path={{ texLive_dest }} state=directory mode=g+rwxs
   tags: ngi_reports
 
-- name: Download TeXLive
+- name: Download TeXLive (Skipped if installed)
   get_url:
     url: "{{ texLive_url }}"
-    dest: "{{ texLive_dest }}/{{ texLive_file }}"
+    dest: "{{ texLive_dest }}/src/{{ texLive_file }}"
+    when: not texlive_installed
```
johanherman: I don't think you would have to worry about this, as the `force` argument to `get_url` is by default `no`. I.e. the file will only be downloaded if the destination does not exist. It is a good solution with the `stat` on L5 otherwise.

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -2,29 +2,44 @@
...
   tags: ngi_reports
 
 - name: Unpack TeXLive
   unarchive:
-    src: "{{ texLive_dest }}/{{ texLive_file }}"
+    src: "{{ texLive_dest }}/src/{{ texLive_file }}"
     dest: "{{ texLive_dest }}"
     copy: no
-    creates: "{{ texLive_path }}/LICENSE"
+    creates: "{{ texLive_dest }}/src/LICENSE"
```
johanherman: Is really the archive unpacked directly under `{{ texLive_dest }}/src/`? Wouldn't it be something like `{{ texLive_dest }}/src/texBlaBla-Foo/LICENSE`?

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -2,29 +2,44 @@
...
     mode: g+rw
   tags: ngi_reports
 
-- name: Append TexLive to $PATH via sourceme
+- name: Copy TexLive profile
+  template: src="texlive.profile.j2" dest="{{ ngi_pipeline_conf }}/texlive.profile" mode=0775
```
johanherman: Does this file need to be executable? Set it to `0664` otherwise.

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -2,29 +2,44 @@
...
+  tags: ngi_reports 
+
+- name: Install TeXLive (Skipped if installed)
+  shell: "./install-tl --profile={{ ngi_pipeline_conf }}/texlive.profile"
+  args:
+    chdir: "{{ texLive_dest }}"
+  when: not texlive_installed
```
johanherman: Here you could use the `creates` option to `shell` instead.

roles/ngi_reports/templates/make_report.sh.j2
```diff
@@ -0,0 +1,12 @@
+# Function to run Pandoc with the correct parameters
+
+function make_report {
+    for f in "$@"; do
+      fn=${f%.md}
+      echo "Converting ${f}.."
+      PD_DIR=${{ ngi_reports_dest }}/data/pandoc_templates
```
johanherman: I guess you do not want that `$` here.

johanherman:
Looks good from what I can tell otherwise. You mentioned that this the TeX stuff can take a humongous amount of time installing though. Which step/task is it specifically that takes up the bulk of that time? Is it those steps where you've been using stat/creates? Just wanna make sure that you're catching it somehow so it will only be executed once and not for every run.

sylvinite:
The download thats a little while, the installation step takes a huge while. Everything else should be ok. I'll reduce the stat to only check the installation step.

johanherman:
:+1:

PR #92, opened 2016-05-22
NGI Reports LaTeX installation fix

sylvinite:
Finally created a working hack that handles LaTeX insistence on not naming their releases with version numbers.


roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -30,10 +30,13 @@
   template: src="texlive.profile.j2" dest="{{ ngi_pipeline_conf }}/texlive.profile" mode=0664
   tags: ngi_reports 
 
+- shell: "find {{ texLive_dest }}/src/install* -maxdepth 0 -type d"
```
johanherman: Note that a problem with all these solutions is that if you're running this in the future, and somehow suddenly `http://mirror.ctan.org/systems/texlive/tlnet/{{ texLive_file }}` is bundling a new version, then you'll end up with several folders that match your expression due to the wildcard. 

If it is good enough to catch the latest created folder under the `src/` subdir, then perhaps you could do something like `shell: ls -1 -td {{ texLive_dest }}/src/install* | head -1`

sylvinite: Did something similar which exploits the naming convention rather than modification date. The LaTeX installation process is awful. Good catch!

roles/ngi_reports/tasks/dependencies.yml
```diff
@@ -18,25 +18,25 @@
   tags: ngi_reports
 
 - name: Unpack TeXLive
-  unarchive:
-    src: "{{ texLive_dest }}/src/{{ texLive_file }}"
-    dest: "{{ texLive_dest }}/src/"
-    copy: no
-    creates: "{{ texLive_dest }}/src/install-tl-*/LICENSE.TL"
-    mode: g+rw
+  command: "tar -xzf {{ texLive_dest }}/src/{{ texLive_file }} -C {{ texLive_dest }}/src/ --strip-components=1"
```
johanherman: Are you trying to unpack everything directly under `{{ texLive_dest }}/src/`? Because on L37 you're launching `install-tl` directly from this dir.

johanherman: Note also that if you're using the `command` instead of the `shell` module then I usually specify the whole path to the binary, as it won't generally have the binary in the `$PATH`.

sylvinite: Yea, everything is unpacked directly under {{ texLive_dest }}/src/ .
Version management under /src/ is pointless. Strip components moves the output.
I could make another folder to pack into (which would be another 3 lines), but it feels a bit silly.

PR #95, opened 2016-05-24
MultiQC source & taca config fix

sylvinite:
Taca deliveries had a folder mismatch.
MultiQC was requested to be pulled from source. As such, script has been altered to reflect this

### DIFF
--- a/roles/multiqc/tasks/main.yml
+++ b/roles/multiqc/tasks/main.yml
@@ -1,7 +1,15 @@
 ---
+- name: Fetch MultiQC from GitHub
+  git: repo="{{ multiqc_repo }}"
+       dest="{{ multiqc_dest }}"
+       version="{{ multiqc_version }}"
+       force=yes
+  tags: multiqc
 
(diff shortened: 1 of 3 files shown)


roles/multiqc/tasks/main.yml
```diff
@@ -1,7 +1,20 @@
 ---
+- name: Fetch MultiQC from GitHub
+  git: repo="{{ multiqc_repo }}"
+       dest="{{ multiqc_dest }}"
+       version="{{ multiqc_version }}"
+       force=yes
+  tags: multiqc
+
+
+#- name: Install MultiQC from pip
```
johanherman: Remove this if we're going to fetch from Github going forth -- or are we not decided?

sylvinite: It's a bit back and forth. To quote the developer he'll be implementing a bunch of small fixes for the Irma env, so using pip is a bit too rigid. Ideally we'll go back to pip at some point.

PR #97, opened 2016-06-03
Install some CPAN modules manually via cpanm 

johanherman:
Needed so we can set some appropriate env vars ($PATH) before running cpanm. By removing anaconda from $PATH cpanm will not pick up the version of libxml2.so that it bundles, which otherwise conflicts with the system installed version under /usr/lib64.

### DIFF
--- a/roles/arteria-siswrap-ws/tasks/sisyphus.yml
+++ b/roles/arteria-siswrap-ws/tasks/sisyphus.yml
@@ -13,10 +13,10 @@
 - name: install File::NFSLock
   cpanm: name=File::NFSLock locallib="{{ perllib_dest }}"
 
-# Don't run the tests for this, as it seems to complain about a 
-# minor version missmatch for libxml2. 
 - name: install XML::LibXML
-  cpanm: name=XML::LibXML locallib="{{ perllib_dest }}" notest=yes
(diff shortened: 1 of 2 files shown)


roles/arteria-siswrap-ws/tasks/sisyphus.yml
```diff
@@ -13,10 +13,10 @@
 - name: install File::NFSLock
   cpanm: name=File::NFSLock locallib="{{ perllib_dest }}"
 
-# Don't run the tests for this, as it seems to complain about a 
-# minor version missmatch for libxml2. 
 - name: install XML::LibXML
-  cpanm: name=XML::LibXML locallib="{{ perllib_dest }}" notest=yes
+  shell: "PATH={{ cpanm_env }} cpanm -l {{ perllib_dest }} XML::LibXML"
+  args: 
+    creates: "{{ perllib_dest }}/lib/perl5/*/XML/LibXML.pm"
```
sylvinite: Why asterix? Also applicable for line 45.

johanherman: Because I'm lazy. :D I've pushed the proper path now though.

PR #98, opened 2016-06-03
Use proper home in irma3's bashrc 

johanherman:


sylvinite:
The comments on lines 7-21 bring a lot of uncertainty. What is permanent, what is temporary?

johanherman:
The homedir hack is hopefully not needed in the future, so if everything looks OK we can remove it later. 

The aliases would be nice to have working, but I don't know how to solve it at the moment due to the newgrp.

PR #103, opened 2016-06-09
New ngi_reports version, multiqc config separation

sylvinite:
Due to the NGI submodule of MultiQC causing problems for Uppsala, the config for MultiQC has now been separated into two.
NGI Reports has also been updated to fix an invalid libprep issue.

### DIFF
--- a/roles/multiqc/templates/multiqc_config.yml.j2
+++ b/roles/multiqc/templates/multiqc_config.yml.j2
@@ -1,4 +1,4 @@
 ---
 no_version_check: True
 push_statusdb: False
-disable_ngi: False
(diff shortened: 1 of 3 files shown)


roles/multiqc/templates/multiqc_config.yml.j2
```diff
@@ -1,4 +1,4 @@
 ---
 no_version_check: True
 push_statusdb: False
-disable_ngi: False
+"disable_ngi: {{ item.ngi_hooks }}"
```
johanherman: Remove the "" if you do not want them to exist in the outputed config file. They are not parsed/required by jira in this context.

PR #104, opened 2016-06-09
NGI-RNAseq role/recipe

sylvinite:
Wrote a NGI-RNAseq role.

### DIFF
--- a/roles/ngi-rnaseq/tasks/main.yml
+++ b/roles/ngi-rnaseq/tasks/main.yml
@@ -1,0 +1,42 @@
+---
+
+- name: Download & install NextFlow 
+  shell: "curl -fsSL get.nextflow.io | bash"
+  args:
+    chmod: "{{ root_path }}/sw"
+  tags: ngi-rnaseq
+
+- name: Add nextflow to $PATH
+  lineinfile: dest="{{ ngi_pipeline_conf }}/{{ bash_env_script }}"
+              line="export PATH={{ root_path }}/sw/nextflow:$PATH"
+              backup=no
+  tags: ngi-rnaseq
+
+- name: Fetch NGI-RNAseq from GitHub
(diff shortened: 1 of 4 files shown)


roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
+---
+
+- name: Download & install NextFlow 
+  shell: "curl -fsSL get.nextflow.io | bash"
+  args:
+    chmod: "{{ root_path }}/sw"
```
johanherman: `chdir`, not `chmod`. 

What do you think about putting the binary in a subdir instead, and not directly under `sw/`?

roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  tags: ngi-rnaseq
+
+- name: Add nextflow to $PATH
+  lineinfile: dest="{{ ngi_pipeline_conf }}/{{ bash_env_script }}"
+              line="export PATH={{ root_path }}/sw/nextflow:$PATH"
```
johanherman: This export is only correct if you put the binary in a subdir called `nextflow`, as I contemplated above.

roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -1,9 +1,13 @@
 ---
+- name: Create nextflow directory
+  file: path="{{ nextflow_dest }}" state=directory mode=g+s
```
johanherman: You can set it to `g+rwXs` while you're at it.

roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -1,9 +1,13 @@
...
+  mode: g+rw
```
johanherman: And remove this `mode`.

PR #106, opened 2016-06-13
Sync .git dirs due to dependencies that need to read it

johanherman:
Sisyphus (that is used by some of the Arteria web services) requires the .git dir to exist in the installed folder, so that it e.g. can read out version information for tracability reasons, etc. This is the easiest fix for this.


johanherman:
This on the other hand highlights the fact that a lot of the files under the .git dirs have wrong permissions (missing g+w).

sylvinite:
Looks alright.

PR #107, opened 2016-06-14
Playbook makeover (file perms and tags) 

johanherman:
- Set tags on role level directly from the playbook instead of on each task 
- Bump installed version of arteria-siswrap
- Install Arteria in main playbook instead of a separate playbook 
- Enforce file perms for all files and dirs under `/lupus/ngi/{sw,conf,resources}` via a playbook level task. Sets the perms only if the user running the playbook owns the files.


sylvinite:
Looks great, really don't have anything to comment on.

PR #108, opened 2016-06-16
Config updates for several software

sylvinite:
- TACA softlinks moved due to multi-softlinks problem
- ngi-rnaseq config and version updated 
- ngi reports config separated into two; hopefully works in both instances
- taca config updated


roles/ngi_reports/tasks/main.yml
```diff
@@ -30,7 +30,7 @@
     site: "upps"
 
 - name: Create Uppsala ngi_reports conf file
-  template: src="ngi_reports_site.conf.j2" dest="{{ ngi_pipeline_conf }}/ngi_reports_{{ site }}.conf"
+  template: src="ngi_reports_{{ site }}.conf.j2" dest="{{ ngi_pipeline_conf }}/ngi_reports_{{ site }}.conf"
```
johanherman: Perhaps we could redo these 2+2 tasks (`template`+`lineinfile`) to use the `with-items` paradigm you've used previously, instead of having `set_fact` two times.

johanherman:
:+1:

PR #110, opened 2016-06-21
Tarzan: use self signed SSL cert valid for 4 years

johanherman:
Today we're using an auto-generated self signed SSL cert located at `/proj/ngi2016001/private/ngi_pipeline/log/tarzan/ssl`. This causes clients to warn about it being self signed. To circumvent this one can add the contents of the cert in the client's CA bundle (e.g. in the one included in Python's `request` library, which is used by Stackstorm). 

But apparently these auto-generated certs are only valid for 1 month, so the client is still complaining as the current cert has expired. 

Therefore we're now instead manually generating a self-signed cert that we put in the Tarzan conf dir.

### DIFF
--- a/roles/tarzan/README.md
+++ b/roles/tarzan/README.md
@@ -19,6 +19,12 @@ There is no specific backup of Cassandra at the moment. And the serf traffic (ov
 
 # Usage 
 
+## Configuring self signed SSL cert
+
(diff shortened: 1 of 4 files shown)


sylvinite:
Looks good.

PR #111, opened 2016-06-21
Minor code structure cleanup

sylvinite:
Cleaned up a few things noticed during development.
Do not merge yet as ngi-rnaseq had some kind of dependency issue I haven't worked out yet.


roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -18,20 +18,15 @@
        version="{{ ngi_rnaseq_version }}"
        force=yes
 
-- name: Install NGI-RNAseq
```
johanherman: This is not required any longer?

sylvinite: It was never required. NGI-RNAseq is a nextflow application.
I figured the first revision of the recipe wouldn't work anyway, so I wasn't too keen on doublechecking everything.

johanherman: :+1:

PR #112, opened 2016-06-22
Versioning draft

sylvinite:
Wrote a short Contributing file and added a version file to more easily keep track of what instance of irma-provisioning is being used.

Feedback is appreciated.


johanherman:
Do you prefer to use a `version.txt` instead of the tags from the `.git` subdir? Note that as it is now `CONTRIBUTING.md` and `version.txt` will not either be synced over to the cluster.

PR #114, opened 2016-06-27
Fix template syntax error

johanherman:
Kong's template uses an other Jinja pattern `[% %]` due to the embedded nginx conf parses `{{ }}` as well.


sylvinite:
Checked the script for any other errors and found none. Looks plenty fine to me.

PR #116, opened 2016-06-30
Fixed ngi-rnaseq recipe

sylvinite:
Apparently the default behaviour is for Nextflow to create a checksummed java arguments file in a folder based on the host name. Due to the way we deploy this behaviour causes some issues. This fix should solve this.


roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -20,6 +15,14 @@
               line="export PATH={{ root_path }}/sw/nextflow:$PATH"
               backup=no
 
+- name: Point NextFlow dependencies vars to installation directory
+  lineinfile: dest="{{ ngi_pipeline_conf }}/{{ bash_env_script }}"
+              line="export { item }={{ nextflow_dest }}"
```
johanherman: `{{ }}` instead of `{ }`.

roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -2,16 +2,11 @@
...
-  shell: "export NXF_HOME={{ nextflow_dest }}"
-
-- name: Add NextFlow installation files to env var NFX_HOME
-  lineinfile: dest="{{ ngi_pipeline_conf }}/{{ bash_env_script }}"
-              line="export NXF_HOME={{ nextflow_dest }}"
-              backup=no
-
 - name: Install NextFlow
   shell: "./nextflow"
+  environment:
```
johanherman: Using `environment` will set these environment variables /during/ the execution of the playbook. Why is this necessary here? I can't see you referencing these anywhere, nor any steps where they will be implicitly picked up by some step.

sylvinite: They are necessary for the './nextflow' command to function correctly; e.g. install files in the directory I want them to be placed in. This is implied by having them set in the 'Install NextFlow' task.

johanherman: Ah! Sorry, my bad. I misread it as `- environment:`. Didn't notice that it was part of the `shell` task, thought it was standalone.

johanherman:
Is the nougat commit supposed to be included in this PR?

PR #118, opened 2016-07-04
Nougat recipe & rna-seq fix

sylvinite:
Nougat recipe that is fully functional. Probably requires a few tweaks before running to 100%.

Also includes a small fix that should (finally) let nextflow for rna-seq run as intended.


roles/nougat/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+       force=yes
+
+- name: Install NouGAT's requirements
+  shell: "{{ ngi_pipeline_venv }}/bin/pip install -r {{ nougat_dest }}/requirements.txt"
+
+- name: Copy configuration
+  template: src="NouGAT_config.conf.j2" dest="{{ ngi_pipeline_conf }}/NouGAT_config.conf"
+
+- name: Install NouGAT
+  shell: "{{ ngi_pipeline_venv }}/bin/pip install -e ."
```
johanherman: Is `-e` needed here as well?

sylvinite: Yea, it's necessary for the entry_points mentioned here https://github.com/SciLifeLab/NouGAT/blob/master/setup.py

johanherman:
:+1:

PR #120, opened 2016-07-06
Updated taca version, created additional delivery configs

sylvinite:
Version update most notable sets a better sender for TACA messages.
Extra delivery configs to easily deliver fastq files to irma.

### DIFF
--- a/roles/taca/templates/sthlm_fastq_delivery.yml.j2
+++ b/roles/taca/templates/sthlm_fastq_delivery.yml.j2
@@ -6,7 +6,7 @@ deliver:
     analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
     datapath: <ROOTPATH>/DATA/<PROJECTID>
     stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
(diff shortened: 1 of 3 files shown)


roles/taca/templates/sthlm_fastq_delivery.yml.j2
```diff
@@ -6,7 +6,7 @@ deliver:
     analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
     datapath: <ROOTPATH>/DATA/<PROJECTID>
     stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
-    deliverypath: /pica/sw/links/proj/<UPPNEXID>/INBOX/<PROJECTID>
+    deliverypath: "{{ item.proj_path }}/<UPPNEXID>/INBOX/<PROJECTID>"
```
johanherman: Don't use quotes (`" "`) if you don't want them in the generated template. It will not be parsed by Jinja.

sylvinite: It's fine either way. Afaik the path is read a string.

PR #126, opened 2016-07-12
WIP: Preparing for parallel deployments

johanherman:
This is work in progress. Do not merge. Mainly showing this to you @sylvinite at this stage because you might have some input before I continue further. 

Supports production/staging/devel areas, with multiple deployment versions within each area.

Still left to do is updating/checking all roles so that e.g. the config files gets correct, and setting up areas where programs can write test data.


tasks/pre-install.yml
```diff
@@ -0,0 +1,38 @@
+--- 
+
+  # These tasks need to run before any other roles, to setup proper 
+  # environment variables depending on we're deploying to production, 
+  # staging, or development/test purposes. 
+  
+  # TODO: Perhaps this logic can be made more pretty somehow. 
+
+  - fail: msg="You must specify '-e deployment_environment=<production|staging|devel>'"
```
sylvinite: This same forceful behaviour can be achieved by using inventory files, which allows the variables to be set there.
It seperates it a bit better rather than writing all three instances in the same file.

sylvinite:
Looks like a great first draft!
I really only had one comment about the structure :+1:

tasks/pre-install.yml
johanherman: You mean changing install.yml from having `hosts: 127.0.0.1` to e.g. `hosts: irma`, and then adding three files 

1) `inventory/production.yml` that reads 

```
[irma]
127.0.0.1 
[irma:vars]
deployment_environment=production
deployment_environment=default
deployment_override=false
```

2)  `inventory/staging.yml` that reads 

```
[irma]
127.0.0.1 
[irma:vars]
deployment_environment=staging
deployment_environment=default
deployment_override=false
```

3) `inventory/devel.yml` that reads 

```
[irma]
127.0.0.1 
[irma:vars]
deployment_environment=devel
deployment_environment=default
deployment_override=false
```

and then forcing the user to run with `ansible-playbook -i inventory/staging.yml install.yml` instead of having `-e deployment_environment=staging`? 

Or perhaps better to then also have a `group_vars/all` that includes 

```
deployment_environment=default
deployment_override=false
```

and letting the inventory only carry the `deployment_environment` variable. Or perhaps then putting this one in `host_vars/127.0.0.1/main.yml`, or moving the contents in that one to `group_vars/all` -- so we keep all generic main variables in one place. 

This will still require the user to pass in the `-e deployment_version` (and perhaps also `-e deployment_override`) though. And I would still prefer to keep the `fail` task on L9 in `pre-install.yml` because an user can still override the values in the inventory by passing in a variable explicitly on the command line, and I would prefer to disallow that, so that the user doesn't accidentally deploy to some weird path. 

A disadvantage with all these separations, and one of the reasons why I put the variables directly in `install.yml` from the beginning is that I want these variables to stand out more clearly, as they are on a kind of meta level compared to the other variables we've got. But perhaps I'm mistaken.

sylvinite:
Reviewed it and it looks great.
I am however a bit hesitant to merging it right now since I'll be going on vacation in a week, and deploying this thing whilst minimizing production downtime will probably require a fair bit of collaborative effort.

johanherman:
:+1: We'll wait until you're back.

sylvinite:
Following extensive discussion and improvements, this PR should be ready to be merged.

PR #128, opened 2016-07-18
Added MultiQC log extension clean patterns

ewels:
These are specific to NGI pipelines, so shouldn't be added to core MultiQC. So this is the place to put them.


sylvinite:
Looks good.

PR #130, opened 2016-08-02
Disable retry files that clutter working area

johanherman:
This will disable the retry files that are usually created with the same prefix as the playbook. It will clean up the working area a little bit as we're not using this feature at the moment.


johanherman:
@sylvinite this is a small thing

sylvinite:
Cute!

PR #132, opened 2016-09-12
Use Piper from module system 

johanherman:
This PR asks ngipipeline to use Piper from the module system instead of deploying it ourselves. It also uses the gatk_bundle provided by UPPMAX. 

It has been test deployed under `devel-root` on `irma3` without problems, but no scripts or programs have been tried. There are some values that might need adjustment after more proper testing.


sylvinite:
Looks good. Uncertain about the setupfilecreator part but lets test it and hope for the best!

PR #137, opened 2016-09-20
Remove piper refs

johanherman:
These things haven't been used lately, so no need to test it in staging. Just cleaning up the repo a little bit before releasing a new version.


sylvinite:
Good stuff

PR #141, opened 2016-09-20
Add PyVCF for joint calling

johanherman:


johanherman:
Self merging this as it was requested by Isak and worked OK to install in devel.

PR #143, opened 2016-09-21
Remove reports stats cronjob for Uppsala 

johanherman:
These were relevant on Nestor; an other solution is being implemented for Irma.


sylvinite:
As long as it works for you guys; I see no problem with this.

PR #144, opened 2016-09-21
Fixes and updates to the staging env 

johanherman:
This PR mainly updates the Arteria roles so that they use the Irma staging/prod environments instead of having their own internal staging/prod combos. 

It also fixes a bug with `tasks/pre-install.yml` where `proj_root` got set to `/proj` even in staging due to a mistake I made when I added the `set_fact` for `milou_delivery`. I assumed the tasks were in a `block` only evaluated for production setups, but instead it got set all the time. 

Then we're now setting the Bash path explicitly instead of re-using the `PS1` variable, because that didn't work when having a staging env as a personal user (although it did work as a func user in production mode). Having it set like this instead makes it work in both setups though.

### DIFF
--- a/roles/arteria-checksum-ws/tasks/main.yml
+++ b/roles/arteria-checksum-ws/tasks/main.yml
@@ -5,24 +5,18 @@
 # Production 
 - set_fact: 
     arteria_checksum_monitored_path: "{{ arteria_checksum_monitored_path }}"
-    arteria_checksum_env_root: "{{ arteria_checksum_env_root }}/production/"
-    arteria_checksum_src_path: "{{ arteria_checksum_src_path }}/production/"
-    arteria_checksum_config_root: "{{ arteria_checksum_config_root }}/production/"
     arteria_checksum_md5sum_log_dir: "{{ arteria_checksum_md5sum_log_dir }}/production"
     arteria_checksum_service_log: "{{ arteria_checksum_service_log}}/production"
     arteria_checksum_port: "{{ arteria_checksum_port_prod }}"
-  when: arteria_checksum_environment == "production"
+  when: deployment_environment == "production"
 
 # Staging
 - set_fact:
     arteria_checksum_monitored_path: "{{ arteria_checksum_monitored_path }}/staging/"
-    arteria_checksum_env_root: "{{ arteria_checksum_env_root }}/staging/"
-    arteria_checksum_src_path: "{{ arteria_checksum_src_path }}/staging/"
-    arteria_checksum_config_root: "{{ arteria_checksum_config_root }}/staging/"
     arteria_checksum_md5sum_log_dir: "{{ arteria_checksum_md5sum_log_dir }}/staging"
     arteria_checksum_service_log: "{{ arteria_checksum_service_log}}/staging"
     arteria_checksum_port: "{{ arteria_checksum_port_stage }}"
-  when: arteria_checksum_environment == "staging"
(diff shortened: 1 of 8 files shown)


roles/arteria-siswrap-ws/defaults/main.yml
```diff
@@ -1,26 +1,33 @@
...
+# staging sync. 
 arteria_siswrap_env_root: "{{ arteria_install_path }}/siswrap_venv"
 arteria_siswrap_sources_path: "{{ arteria_install_path }}/siswrap_src"
 arteria_siswrap_config_root: "{{ ngi_pipeline_conf }}/arteria/siswrap"
 arteria_siswrap_app_config: "{{ arteria_siswrap_config_root }}/app.config"
 arteria_siswrap_logger_config: "{{ arteria_siswrap_config_root }}/logger.config"
-arteria_siswrap_log: "{{ ngi_pipeline_upps_path }}/log/arteria/siswrap-ws"
-runfolder_path: "{{ proj_root }}/{{ ngi_pipeline_upps_delivery }}/incoming/"
+arteria_siswrap_log: "/proj/{{ ngi_pipeline_upps_delivery }}/private/log/arteria/siswrap-ws"
+runfolder_path: "/proj/{{ ngi_pipeline_upps_delivery }}/incoming/"
```
sylvinite: This and a few other instances always pumps the output to /proj/ even for a staging instance. Is this intended?

roles/arteria-checksum-ws/tasks/main.yml
```diff
@@ -5,24 +5,18 @@
...
 - set_fact:
     arteria_checksum_monitored_path: "{{ arteria_checksum_monitored_path }}/staging/"
-    arteria_checksum_env_root: "{{ arteria_checksum_env_root }}/staging/"
-    arteria_checksum_src_path: "{{ arteria_checksum_src_path }}/staging/"
-    arteria_checksum_config_root: "{{ arteria_checksum_config_root }}/staging/"
     arteria_checksum_md5sum_log_dir: "{{ arteria_checksum_md5sum_log_dir }}/staging"
     arteria_checksum_service_log: "{{ arteria_checksum_service_log}}/staging"
     arteria_checksum_port: "{{ arteria_checksum_port_stage }}"
-  when: arteria_checksum_environment == "staging"
+  when: deployment_environment in [ "staging", "devel" ]
```
sylvinite: devel?

roles/arteria-siswrap-ws/defaults/main.yml
johanherman: Yeah. Especially the `runfolder_path` and the `arteria_checksum_monitored_path`. The Arteria staging services are tested in our complete workflow, i.e. including things that are happening automatically here in Uppsala. So our services here need to be able to upload data to the service at UPPMAX. And that can only happen inside the incoming dir under `/proj`. 

For the logs dir... that's more a tracability issue. Because `wildwest` is cleaned with every sync of the staging environment. But as these are long running services then it is nice to be able to go back and look at historical events, therefore I changed them to be put under `/proj` as well so they don't get removed.

roles/arteria-checksum-ws/tasks/main.yml
johanherman: Yes. `devel` is the value of `deployment_environment` if you're running `ansible-playbook` on irma3 without specifying the `-e` flag to override it. I.e. the value that gets set when you're inside `devel/{isak,johanhe}/` and testing that the playbook works.

PR #146, opened 2016-09-22
Fix to add piper module to sbatch

sylvinite:
This is verified on milou to properly add piper to the sbatch instead of it relying on it being loaded by the user. Props to Denis.


johanherman:
Sweet!

PR #147, opened 2016-09-27
Added a few dirs to be created

sylvinite:
If ngi_pipeline is ran with a relative flowcell path it checks out the `incoming` directory. As such we should ensure that there always exists one.
Having the recipe create a `denovo` directory for denovo QC rather than relying on a nice user creating one is also a good idea.


roles/ngi_pipeline/tasks/main.yml
```diff
@@ -32,6 +32,13 @@
 - name: Create ngi pipeline conf directory 
   file: path="{{ ngi_pipeline_conf }}" state=directory mode=g+s
 
+- name: Create ngi pipeline incoming directories
```
johanherman: I presume this is relevant mostly for staging envs?

PR #151, opened 2016-10-04
Create staging links to validation data

johanherman:
`/lupus/ngi/validation_data/` is manually populated with validation data on Irma, but not on irma3. Because `/lupus/ngi/staging/wildwest` is cleaned with every staging sync it is not enough to have static symlinks setup, but we need to re-create them for every deployment. 

A potential problem with this might come up in the future when rsyncing (with the `--delete` flag) as an other user, and if some of these symlinks are supposed to be removed/renamed. Then the rsync might fail with this, as file permissions and symlinks are a different kind of beast. In that case the symlinks will have to be removed manually by the file owner.

### DIFF
--- a/host_vars/127.0.0.1/main.yml
+++ b/host_vars/127.0.0.1/main.yml
@@ -49,5 +49,10 @@ statusdb_creds: statusdb_creds.yml
 
 recipient_mail: ngi_pipeline_operators@scilifelab.se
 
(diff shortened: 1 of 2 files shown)


sylvinite:
Looks good. Love the solution for not being able to create the incoming dir (which we discussed earlier).

PR #153, opened 2016-10-07
Add a bootstrap script and make the README a bit more clear

johanherman:
Bit of a cleanup to the README. Also added a very simplistic bootstrap bash script that can be run the first time instead of having to guess what to do from the previous README instructions.


sylvinite:
Objective improvement.

PR #154, opened 2016-10-11
Provision the GATK key

johanherman:
When we removed the piper role (and started using piper from the module system) we also stopped provisioning the GATK key, because that was a task in the role. Therefore I'm now adding the task to the ngi_pipeline role instead.


sylvinite:
Nothing to complain about.

PR #159, opened 2016-10-14
Wildwest is now be globally editable

sylvinite:
Not assigning anyone since I need to triple-check the logic. It works, but it should also make sense.

### DIFF
--- a/tasks/post-install.yml
+++ b/tasks/post-install.yml
@@ -23,7 +23,7 @@
   - block:
 
     - name: Make wildwest directory editable by anyone that can access irma
-      shell: "find {{ proj_root }} -user `whoami` -exec chmod g+rwX,o=rwX {} \\;"
+      shell: "find {{ proj_root }} -user `whoami` -exec chmod g+rwXs,o=rwX {} \\;"
     when: deployment_environment == "staging"
 
   - debug: msg="Finished deploying to {{ root_path }}"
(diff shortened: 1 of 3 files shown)


sylvinite:
Should be good now

tasks/pre-install.yml
```diff
@@ -61,7 +61,7 @@
         incoming_dir_upps: "{{ proj_root }}/{{ ngi_pipeline_upps_delivery }}/incoming/"
         incoming_dir_sthlm: "{{ proj_root }}/{{ ngi_pipeline_sthlm_delivery }}/incoming/"
     - name: create staging proj root dir (wildwest), milou-delivery and incoming dirs 
-      file: path={{ item }} state=directory mode="g+rwxs,u+rwx" group="ngi"
+      file: path={{ item }} state=directory mode="g+rwx,u+rwx"
```
johanherman: Some specific reason why you removed the setgid flag?

tasks/post-install.yml
```diff
@@ -20,5 +20,11 @@
       file: dest="/lupus/ngi/{{ deployment_environment }}/latest" src="{{ root_path }}" state=link
     when: deployment_environment != "devel" 
 
+  - block:
+
+    - name: Make wildwest directory editable by anyone that can access irma
+      shell: "find {{ proj_root }} -user `whoami` -exec chmod g+rwX,o=rwX {} \\;"
```
johanherman: It could be good to have the setgid flag set here as well.

PR #160, opened 2016-10-19
New TACA version, dependencies and staging_statusdb config

sylvinite:
New TACA version for the cleanup command. Also needed a TACA config, which in turn required a new version of statusdb. TACA github repo will be changed after successful testing.
Additionally made it so that staging versions use the tools-dev server rather than tools.


.gitignore
```diff
@@ -2,6 +2,7 @@
 *.retry
 pontus.larsson_medsci.uu.se.key
 charon_credentials.yml
-statusdb_creds.yml
+statusdb_creds_prod.yml
```
sylvinite: After discussion we concluded that this is fine.
But yes, it needs a status_db credentials file.

roles/taca/templates/taca_site_config.yml.j2
```diff
@@ -0,0 +1,16 @@
+log:
```
johanherman: Shouldn't you use this template in some task somewhere?

roles/taca/templates/taca_site_config.yml.j2
```diff
@@ -0,0 +1,16 @@
...
+    log_file: "{{ ngi_pipeline_site_path }}/taca.log"
+
+cleanup:
+    irma:
+        flowcell:
+            root: 
+                - "{{ proj_root }}/{{ ngi_pipeline_sthlm_delivery }}/incoming"
```
johanherman: Should it be `sthlm`?

.gitignore
johanherman: @sylvinite shouldn't you change https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/0523da869b8a9c02e5014e9d25ec7bdfb470b8d4/roles/ngi_reports/tasks/dependencies.yml#L30 and https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/121c1e2c2fb1086e05994c0cb7c32c791a91db79/roles/multiqc/tasks/main.yml#L28 to read `{{ ngi_pipeline_conf }}/statusdb_creds.yml` (or something) instead of `{{ ngi_pipeline_conf }}/{{ statusdb_creds }}`. I guess it would make things easier downstreams.

sylvinite: I wrote it so that {{ statusdb_creds }} is set to point to either statusdb_creds_prod.yml or statusdb_creds_stage.yml. I think that should work.

roles/taca/templates/taca_site_config.yml.j2
sylvinite: If you're referring to "site" I figured I'd make the template generic then just omit the uppsala creation.

roles/taca/templates/taca_site_config.yml.j2
sylvinite: Yea... That's something I should've added.

roles/taca/templates/taca_site_config.yml.j2
johanherman: Ok.

roles/taca/templates/taca_site_config.yml.j2
johanherman: :+1:

.gitignore
johanherman: No. That's why I'm asking. Because then the file will be named differently when actually deployed on the cluster. And then the software reading this file will have to know the name of it, and that it will change (and to what). But if it is enough with having that environment variable set then I guess there are no worries.

PR #163, opened 2016-11-24
Updated taca & taca-ngi, reverted ngi_pipeline to safe version

sylvinite:
Interest in deploying other stuff before ngi_pipeline (again).
Reverted ngi_pipeline to official version.
Updated taca-ngi & TACA to allow specific cluster delivery.

### DIFF
--- a/host_vars/127.0.0.1/main.yml
+++ b/host_vars/127.0.0.1/main.yml
@@ -16,9 +16,9 @@ bash_env_script: sourceme_common.sh
 bash_env_upps_script: sourceme_upps.sh
 bash_env_sthlm_script: sourceme_sthlm.sh
 
-ngi_pipeline_repo: https://github.com/sylvinite/ngi_pipeline.git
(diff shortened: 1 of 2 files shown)


johanherman:
No updated configs required? Only new command line options?

sylvinite:
Only config difference will be a delivery script/config. But to paraphrase the author it is easier that we just create one locally and add a formal one later to the repo.

johanherman:
Create one locally? Where? On irma3 that is not checked in, or somewhere on the cluster?

sylvinite:
Locally for the user.
Creating the delivery yaml on irma3 and rsyncing it without checking it in would be more complicated than including it.
If I intended to create a file on the cluster I'd append locally with 'on the cluster'.

johanherman:
Well, "locally for the user" must obviously be "locally for the user on the cluster" then, as there can't be many other options. Why I'm asking is because we want the environment to be able to be bootstrapped from scratch at all times from irma-provision on irma3, without having to trust that undocumented magical files will exist here and there. But if the config is supposed to be added to the repo later then go for it.

PR #164, opened 2016-11-24
UPPMAX has changed the 2f challenge

johanherman:
So we need to update the sync script so that pexpect knows what to wait for before sending the user's token.


sylvinite:
Golden!

PR #166, opened 2016-11-28
removed R from modules to load

b97pla:
Loading R caused a reload of java, switching version from 1.7 to 1.8

### DIFF
--- a/roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2
+++ b/roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2
@@ -25,7 +25,6 @@ piper:
     #    required_autosomal_coverage: 28.4
     load_modules:
         - bioinfo-tools piper/{{ piper_module_version }}
-        - R/3.2.3
     threads: 16
     job_walltime:
         merge_process_variantcall: "10-00:00:00"


b97pla:
@vezzi @sylvinite @johanherman

johanherman:
@b97pla we don't need R for some plots or reports or something?

vezzi:
some comment for me, are we sure R is not needed... maybe by GATK

sylvinite:
I have no idea what the usage for the R module is/was for either. 
Doing a quick scan of `ngi_pipeline` yields me nothing.
I do suspect the module is mounted in order to follow http://gatkforums.broadinstitute.org/wdl/discussion/2899/howto-install-all-software-packages-required-to-follow-the-gatk-best-practices .

sylvinite:
Also asked Denis about this:
"R is required for a lot of analysis steps. NGI pipeline doesn't necessarily care about R but I would bet piper does. The upcoming RNA-seq pipeline also uses R in several steps."

johanherman:
From what I've heard we're not using the reports that piper can generate which requires R. @b97pla can confirm.

b97pla:
yes, that is what dahlberg claims and we don't use it. There are other steps that can generate plots, e.g. BQSR, but we don't use that either. I haven't tested though, so I can't guarantee anything.

sylvinite:
I'll go ahead and merge this and test it together with the ngi_pipeline PRs. If it works it works. Worst case scenario I'll just roll back this line for the next production release.

PR #168, opened 2016-12-05
Load PDC archiving settings on login

johanherman:
Use different config files for archiving to PDC depending on if we're logged in as Uppsala or Stockholm.


sylvinite:
Wouldn't you want to set these in default or host_vars?

johanherman:
Yeah. I was lazy. :grin: Fixed now.

PR #169, opened 2016-12-16
Add uppnex_datadeliver script to be used by funk_004

johanherman:
Script that is used by Uppsala for making manual deliveries into INBOXes. Adds this to our PATH.

### DIFF
--- a/host_vars/127.0.0.1/main.yml
+++ b/host_vars/127.0.0.1/main.yml
@@ -1,10 +1,7 @@
 ---
-# For now, values here should be updated accordingly when user develops locally. 
+sw_path: "{{ root_path }}/sw/"
 
-# root_path should be changed to suitable value when e.g. deploying locally
-# for a developer
-#root_path: /lupus/ngi/
-anaconda_path: "{{ root_path }}/sw/anaconda"
+anaconda_path: "{{ sw_path }}/anaconda"
 ansible_path: "/lupus/ngi/irma3/ansible-env/" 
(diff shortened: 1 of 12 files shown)


sylvinite:
Looks good, but some questions:

- uppnex_datadeliver is a bash script but lacks suffix?
- Is this necessarily a part of ngi_pipeline?
- {{ root_path }}/sw/ would more intuitively be defined as something generic under host vars?
- Why is {{ root_path }}/sw/ added per site?
- Is this going to be used by stockholm? Else you could move the task a few rows up to implicitly show it's mostly relevant for uppsala.

johanherman:
- Suffixes are pretty much irrelevant. It has been named like this since it started to be used in production by Uppsala on Milou/Nestor a couple of years ago, so to not confuse our bioinformaticians too much it is best to keep it without the suffix. 

- No, it is not necessarily a part of ngi_pipeline. I was considering adding it to the func role first, but we deploy the `sourcme_site` template in the ngi_pipeline so that's why it is more convenient to do it here. 

- Yeah, perhaps we should add a variable for `{{ root_path }}/sw`; it is used like this in many places among our roles. 

- No, it is not supposed to be used by Stockholm. It is only used by Uppsala a couple of months more while we still have the pica mount on irma4; when that disappears a new system for delivering runfolders will be introduced for both Uppsala and Stockholm.

johanherman:
Have added a commit with some changes now.

sylvinite:
Looks good

PR #174, opened 2017-01-18
Bumped ngi_visual repo, pathed standalone_scripts

sylvinite:
New NGI_visualizations version is necessary for some of the newest MultiQC features.

Standalone scripts has been requested to be added to path. Since this repo is way more liberal than others, I am adding it the end of PATH to avoid the scripts taking priority in possible future naming conflicts. This will hopefully _never_ happen, but this is at least a minor safeguard for it.

### DIFF
--- a/roles/standalone_scripts/tasks/main.yml
+++ b/roles/standalone_scripts/tasks/main.yml
@@ -1,7 +1,14 @@
-#Kind of meager recipe, but will most likely grow from dependencies
+# Kind of meager recipe, but will most likely grow from dependencies
 
 - name: Fetch standalone_scripts from GitHub
   git: repo="{{ ss_repo }}"
        dest="{{ ss_dest }}"
        version="{{ ss_version }}"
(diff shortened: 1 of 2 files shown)


roles/standalone_scripts/tasks/main.yml
```diff
@@ -1,7 +1,14 @@
...
 - name: Fetch standalone_scripts from GitHub
   git: repo="{{ ss_repo }}"
        dest="{{ ss_dest }}"
        version="{{ ss_version }}"
        force=yes
+
+# Since the standalone repo is more free-form than others
+# putting PATH at end of $PATH as a safeguard from future problems
+- name: Add standalone scripts to end of $PATH via sourceme
+  lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_script }}
```
johanherman: I would prefer if you only added this to Stockholm's environment, and not globally.

PR #175, opened 2017-02-07
make ngi_reports required for Uppsala deliveries

b97pla:
As the title suggests, deliveries will now fail if ngi_reports fail to be generated, which is the desired behaviour for Uppsala.


sylvinite:
Looks fine by me, @johanherman

PR #176, opened 2017-02-07
Don't add ngi_reports' script dir to $PATH 

johanherman:
Having the scripts dir in $PATH causes name collision in some situations when resolving the script name `ngi_reports`, as the one we want to call is located in Anaconda's venv dir (that is also in $PATH). 

The files under the scripts dir are helpers that are needed for functionality though, but it should apparently have been solved through other ways already. This needs to be tested though for certainty.

### DIFF
--- a/roles/ngi_reports/tasks/main.yml
+++ b/roles/ngi_reports/tasks/main.yml
@@ -16,11 +16,6 @@
   args:
     chdir: "{{ ngi_reports_dest }}"
 
-- name: Add ngi_reports scripts to $PATH via sourceme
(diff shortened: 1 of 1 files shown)


sylvinite:
Pretty sure this will work and that the `scripts` addition to PATH is a remnant from a bygone age. See https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/0523da869b8a9c02e5014e9d25ec7bdfb470b8d4/roles/ngi_reports/tasks/dependencies.yml#L16 where a similar fix is implemented in the recipe.

PR #177, opened 2017-02-07
Snapshot Cassandra weekly

johanherman:
This will cause a snapshot to be taken of the Cassandra db (Kong's backend) once a week, so that the normal backup processes that run on `/proj/` dirs can include the files in a well known state. The created snapshots are so minimal (a couple of hundred kb) so no need to do any clean up at the moment.


johanherman:
Missing one commit. Don't merge yet.

PR #178, opened 2017-02-09
Tweaks, mainly for ngi-rnaseq

sylvinite:
- Fixed a fastq_screen outdated reference
- Removed test_devel role
- Updated ngi-rnaseq & ngi_visualisations versions (for validation), & standalone_scripts
- ngi-rnaseq now uses a new config base, and the alias is reintroduced
- Script now removes softlinks dir when re-run, thus forcing recalculation of md5's (which is good)
- IGN reports were removed from taca delivery for stockholm. Scripts were also sorted a tiny bit
- Anaconda had a checksum issue, so quickly fixed that

I also need to apply a temporary ngi_pipeline branch (for staging), but I figured these changes could be reviewed prior that; since it hardly affects anything from the playbook's point of view.


roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -6,11 +6,21 @@
        force=yes
 
 - name: Copy NGI-RNAseq config
-  template: src="ngi-rnaseq.config.j2" dest="{{ ngi_pipeline_conf }}/ngi-rnaseq_{{ item.site }}.config"
+  template: src="{{ ngi_rnaseq_dest }}/conf/uppmax.config" dest="{{ ngi_pipeline_conf }}/ngi-rnaseq_{{ item.site }}.config"
   with_items:
   - { site: "sthlm", proj: "{{ ngi_pipeline_sthlm_delivery }}" }
```
johanherman: Guess we could remove the `proj` param then.

roles/ngi-rnaseq/tasks/main.yml
```diff
@@ -6,11 +6,21 @@
...
   - { site: "upps", proj: "{{ ngi_pipeline_upps_delivery }}" }
 
+- name: Apply NGI specific config changes
+  lineinfile: dest="{{ ngi_pipeline_conf }}/ngi-rnaseq_{{ item.site }}.config"
```
johanherman: Have you tested this, or is this as expected? Because `lineinfile` will here just ensure that the lines that you specify will be added, and when I looked at https://github.com/SciLifeLab/NGI-RNAseq/blob/master/conf/uppmax.config I was led to believe that the config should have different syntax. But perhaps that is how Nextflow works?

sylvinite: This should work. Even had a discussion if it is ok to place them at the end of the file.

johanherman: :+1:

PR #191, opened 2017-03-01
Updated ngi-rnaseq dependencies, updated taca version

sylvinite:
**ngi-rnaseq**
Cleared unnecessary config sections
Added R dependency download
Removed NXF_WORK possibly being under /scratch
**taca**
Version update for some cleanup procedures

### DIFF
--- a/roles/ngi-rnaseq/tasks/dependencies.yml
+++ b/roles/ngi-rnaseq/tasks/dependencies.yml
@@ -7,8 +7,11 @@
 - name: Create NextFlow directory
   file: path="{{ nextflow_dest }}" state=directory mode=g+rwXs
 
-- name: Create NextFlow work directory
-  file: path="{{ nextflow_dest }}/workfiles" state=directory mode=g+rwXs group="ngi"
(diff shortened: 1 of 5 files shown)


roles/ngi-rnaseq/tasks/dependencies.yml
```diff
@@ -44,7 +47,25 @@
...
   - { envvar: "export NXF_LAUNCHBASE=${SNIC_TMP:-{{ nextflow_dest }}/workfiles}"}
   - { envvar: "export NXF_LAUNCHER=${SNIC_TMP:-{{ nextflow_dest }}/workfiles}"}
 
+- name: Copy R dependencies script
+  copy:
+    src: "files/install_R_dependencies.r"
+    dest: "{{ nextflow_dest }}/r_packages"
+ 
+- name: Load R module
+  shell: "module load R"
```
johanherman: You know that all these `shell` and `command` calls live in their own separate environments right? They will not know about each other.

johanherman:
:+1:

PR #210, opened 2017-03-31
Addded kmergenie for nougat

sylvinite:
The permissions might need a review since I am unsure about them.
Kmergenie explictly needed `u+x` to even consider running.

### DIFF
--- a/roles/nougat/tasks/dependencies.yml
+++ b/roles/nougat/tasks/dependencies.yml
@@ -43,7 +43,7 @@
   file:
     path: "{{ kmer_dest }}/kmergenie"
     state: touch
-    mode: "u+rwxs"
+    mode: "u+rwx"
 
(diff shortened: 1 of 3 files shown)


roles/nougat/tasks/dependencies.yml
```diff
@@ -16,6 +16,45 @@
...
+- name: Install Kmergenie
+  shell: "make"
+  args:
+    chdir: "{{ kmer_dest }}"
+
+- name: Fix KmerGenie permission issue
+  file:
+    path: "{{ kmer_dest }}/kmergenie"
+    state: touch
+    mode: "u+rwxs"
```
johanherman: You probably don't want to set the binary setuid, as that means the binary will be run with the permission belonging to the file owner (i.e. the user that deployed the sw).

PR #231, opened 2017-05-15
Added acknowledgements file, removed denovo folder creation

sylvinite:


roles/taca/templates/site_fastq_delivery.yml.j2
```diff
@@ -18,6 +18,9 @@ deliver:
     hash_algorithm: md5
     files_to_deliver:
         -
+            - {{ ngi_site_softlinks }}/ACKNOWLEDGEMENTS.txt
+            - <STAGINGPATH>
```
johanherman: Are these two lines as expected?

sylvinite: Yes I believe so. Is there anything odd about them?

johanherman: Aha. Noticed now that it is a nested list, as you've got an other `-` indented differently on line 20.

PR #248, opened 2017-08-23
Workaround for failing PDL dependency tests

johanherman:
An other PDL dependency seems to suddenly fail some of its tests. This workaround just installs the dependency before PDL, and then asks for the tests to be skipped. It might perhaps introduce some weird behaviour downstream (untested), but if that is the case then we'll just have to solve it then.


sylvinite:
:+1:

PR #253, opened 2017-09-19
remove unwanted alias for ngi-methylseq

senthil10:
Only bismark will be used for production, so @ewels prefer just one alias. This is something not urgent so will be included in next release.


ewels:
Thanks! I should really move the bwa-meth stuff out into a subfolder / other repo / just delete it. I'm half waiting for nextflow sub-workflows so that I can keep it there but in a tidy way. One day....

ewels:
Also note the other change that @senthil10 made here - `nextflow run` instead of just `nextflow`. Both work for now but the latter will break in the future.

senthil10:
Great @ewels 👍  and yes this also changes it to `nextflow run`

PR #257, opened 2017-10-09
Increase the performance slightly

johanherman:
This will increase the performance of Ansible slightly. As default it only keeps the SSH connections open for 1 minute, so lets increase it to 60 minutes. Pipelining reduces the number of SSH operations required for executing a module. Can cause problems when combined with sudo, but that's not an issue for us.


senthil10:
Nice @johanherman 👍

PR #258, opened 2017-11-06
CAW recipe 

senthil10:
Added new role for `CAW`


roles/caw/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+       dest="{{ caw_dest }}"
+       version="{{ caw_version }}"
+       force=yes
+
+- name: Create CAW Singularity folder
+  file: name="{{ caw_singularity }}"
+        state=directory mode=g+s
+
+- name: Create temp directory for CAW singularity images
+  command: "mkdir {{ item }}"
```
johanherman: Any specific reason for why you're using `mkdir` as a shell command instead of using `file` with `state=directory`?

roles/caw/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+  with_items:
+  - "{{ caw_singularity }}/singularity_cache"
+  - "{{ caw_singularity }}/singularity_temp"
+
+- name: Pull CAW Singularity images
+  command: "singularity build {{ item }}-{{ caw_version }}.img {{ caw_singularity_repo }}/{{ item }}:{{ caw_version }}"
+  environment:
+    SINGULARITY_CACHEDIR: "{{ caw_singularity }}/singularity_cache"
+    SINGULARITY_TMPDIR: "{{ caw_singularity }}/singularity_temp"
+  args:
```
johanherman: Perhaps a `creates` can be added here if you know how the output of the path for the build will look like? Or - if we do not care about keeping several versions of each container, just point look for the directory path that is used instead?

roles/caw/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+    chdir: "{{ caw_singularity }}"
+  with_items: "{{ caw_singularity_images }}"
+  ignore_errors: true
+
+- name: Remove CAW singularity temp directories
+  command: "rm -rf {{ item }}"
```
johanherman: Considering how many shell commands you're using instead of the `file` module I'm guessing I might be missing something here, but otherwise I would have suggested the `state=absent` argument instead here.

roles/caw/tasks/main.yml
senthil10: Initially I used `/scratch` for `singularity_cache` and `singularity_temp` and also I did not cleanup. Thought `file` module dint have `creates` arg so used `command`.

But later realised it (the cache and temp directories) requires a lot of space and changed their path within new deployment env itself (as I would cleanup them later). Then it did not strike me to use `file` but yes I can change it 👍

roles/caw/tasks/main.yml
senthil10: Actually it should not exists before as the `{{ caw_singularity }}` is created each time for new deployment

roles/caw/tasks/main.yml
senthil10: Again no specific, since I am new to `ansible` I am only using the obvious/simple modules. So `file` with `state=absent` removes the directory ? That is really cool, I will change this as well 👍

roles/caw/tasks/main.yml
senthil10: But yes, you are right. Adding `creates` is not gonna hurt and also might be useful while test deploying

johanherman: Ah, yeah, that will be true in most of the cases, indeed. But doesn't hurt as you say, as it will speed things up e.g re-building, as well making it a bit more future proof.

senthil10:
Tack @johanherman for nice suggestions, I have now included them 👍

PR #260, opened 2017-11-11
Version fix for nextflow and R (ngi-rnaseq)

senthil10:


senthil10:
`self.merging()` as they are pretty straight forward fixes with only version numbers.

PR #261, opened 2017-11-13
Initial role for archive-upload from Irma

johanherman:
Should be mostly similar to what we do locally on biotank, with some differences with the values in the config. This might not work 100% yet though, as it was some time since I last tested this.

### DIFF
--- a/roles/archive-upload-ws/defaults/main.yml
+++ b/roles/archive-upload-ws/defaults/main.yml
@@ -3,14 +3,13 @@
 # If deploying to staging one should call with 
 # -e deployment_environment=staging
 # One can also override the default installed version 
-# by overriding the arteria_checksum_version on the command line. 
+# by overriding the archive_upload_version on the command line. 
 #
 # This will set corresponding paths and use the appropriate port.  
 
 archive_upload_repo: https://github.com/johanherman/snpseq-archive-upload.git
 archive_upload_version: new-service
 
-# {{ ngi_pipeline_upps_delivery }}
 # in production this should be  /proj/ngi2016001/nobackup/NGI/ANALYSIS/<PROJECT ID>
 archive_upload_monitored_path_prod: "/proj/ngi2016001/nobackup/arteria/tmp"
 archive_upload_monitored_path_stage: "/proj/ngi2016001/nobackup/arteria/tmp"
--- a/roles/archive-upload-ws/defaults/main.yml
+++ b/roles/archive-upload-ws/defaults/main.yml
@@ -11,11 +11,11 @@ archive_upload_repo: https://github.com/johanherman/snpseq-archive-upload.git
 archive_upload_version: new-service
 
 # in production this should be  /proj/ngi2016001/nobackup/NGI/ANALYSIS/<PROJECT ID>
-archive_upload_monitored_path_prod: "/proj/ngi2016001/nobackup/arteria/tmp"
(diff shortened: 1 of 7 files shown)


roles/archive-upload-ws/tasks/2_install_config.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: modify uppsala's supervisord conf to start archive-upload-ws 
+  ini_file: 
+    dest="{{ ngi_pipeline_conf }}/supervisord_upps.conf"
+    section="program:archive-upload-ws-{{ deployment_environment }}"
+    option=command
+    value="{{ archive_upload_env_root }}/bin/archive-upload-ws --configroot={{ archive_upload_config_root }} --port={{ archive_upload_port }}"
+    backup=no
+ 
+- name: modify uppsala's supervisord conf to start archive-upload-ws
```
slohse: Is this meant to stay or just for the transition? I recently made some bad experiences with interdependent roles that need to be executed in a certain order... :D

roles/archive-upload-ws/defaults/main.yml
```diff
@@ -0,0 +1,35 @@
...
+
+archive_upload_repo: https://github.com/johanherman/snpseq-archive-upload.git
+archive_upload_version: new-service
+
+# in production this should be  /proj/ngi2016001/nobackup/NGI/ANALYSIS/<PROJECT ID>
+archive_upload_monitored_path_prod: "/proj/ngi2016001/nobackup/arteria/tmp"
+archive_upload_monitored_path_stage: "/proj/ngi2016001/nobackup/arteria/tmp"
+# in production this should be /proj/ngi2016001/nobackup/arteria/pdc_archive_links
+archive_upload_path_to_archive_root_prod: "/proj/ngi2016001/nobackup/arteria/pdc_archive_links/staging/"
+archive_upload_path_to_archive_root_stage: "/proj/ngi2016001/nobackup/arteria/pdc_archive_links/staging/"
```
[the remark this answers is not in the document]
johanherman: 👍

roles/archive-upload-ws/tasks/2_install_config.yml
```diff
@@ -0,0 +1,41 @@
...
+    path: "{{ archive_upload_config_root }}"
+    mode: g+rwXs
+
+#- name: ensure arteria-dsmc log dir exists
+#  file: 
+#    state: directory
+#    path: "{{ arteria_dsmc_log_dir }}"
+#    owner: arteria
+#    group: arteria
+#    mode: g+rwXs
```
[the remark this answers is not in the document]
johanherman: This can be removed. Old stuff from biotank I think. Log dirs can't be created from within the playbook, as those reside under `/proj/ngi2016001` tree.

roles/archive-upload-ws/tasks/2_install_config.yml
johanherman: Yeah. It should, I think. We do this in a bunch of places for Irma, and if I remember correctly it wasn't as easy as just pulling in the upstream role as a dependency, because then that role would override things that OTHER roles have modified. I.e. everything would have to be refactored for that.

johanherman:
Fixed some of your comments.

roles/archive-upload-ws/tasks/2_install_config.yml
slohse: Alright, I suspected something like that :)

slohse:
:+1:

PR #262, opened 2017-11-13
Signal new Irma production envs and automatically restart Uppsala crontabbed services

johanherman:
This PR adds a crontabbed script for Uppsala that every 5 minute will check if the signal file under /tmp/ is present. If it is, then it will kill the defined Uppsala services, reload the new Uppsala crontab file, and then launch the services again. 

For this to work we need to be able to signal when a new production environment has been deployed. At the moment I solve this by modifying the sync script. After the rsync has finished the user will be prompted for a new two factor, and then the script will login and touch the signal file. 

Perhaps this can be done in a better way though. If you have any ideas, please let me know.

### DIFF
--- a/sync.py
+++ b/sync.py
@@ -7,7 +7,7 @@
 #
 # Note that: 
 # 
-# 1. This script assumes that all files to be pupulated are owned by group ngi-sw
+# 1. This script assumes that all files to be populated are owned by group ngi-sw
 # and with appropriate file permissions (group read/write, world read). This should
 # be the case if the deployment bash init file have been sourced before installing sw. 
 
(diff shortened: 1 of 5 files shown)


sync.py
```diff
@@ -20,159 +20,192 @@
...
+		print('Waiting for success')
+		child.expect(exp_success)
+		child.sendline("Logged in with password + factor, logging out.") 
+		child.sendline("exit")
+	elif recv == 1:
+		print('Logged in with password, logging out.')
+		child.sendline("exit")
+
+def signal_new_env(user, host, password, token): 
+	signal_new_env = "/usr/bin/ssh {}@{} 'touch /tmp/.new_irma_env_deployed && chgrp ngi-sw /tmp/.n    ew_irma_env_deployed'".format(user, host)
```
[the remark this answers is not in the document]
johanherman: Indeed, thanks. I'll testrun this a bit more before it is mergable. I just wanted to put it out there when I finally took the time and did this.

johanherman:
Fixed some errors I missed yesterday when I just hacked this together. 

Commented out the specific signaling feature though, as I'd be interested in your input on this first. 

As you say @senthil10 having the services in the crontabs reference `latest` instead of a specific version might make some things easier. 

I think we primarily have used it like this because we wanted to be explicit. You can see from the crontab, and from the currently running processes (`ps` etc) what version is running. Yesterday I thought there were more reasons behind this, but I can't find it documented anywhere, and I can't figure out what it could be - if there are any. So if you and @slohse don't see any problems I'm overlooking at the moment then perhaps I should switch to using `latest` instead. 

But using `latest` is not a silver bullet though. E.g. there still needs to be a signal that tells services to reload. If we e.g. have a crontab job that every minute checks the timestamp of `latest` it could guess that a new environment has been deployed. But when rsync writes the new `latest` it will probably not be done until 10-30 minutes later with the rest of the sync. Therefore the crontabed poller will kill the running services and fail to restart the new services, as they are still not fully deployed. Which means at best there will be a downtime of 10-30 minutes or so. 

So I'm not completely sure which way is best.

senthil10:
@johanherman Yes, I too agree that when ever a new version of production is deployed it is much cleaner/safer to restart the `cronjobs`. 

How about the following approach ? Of course upto you to decide, as we are not running any `cronjob` here 😉  (I should verify that though, why we are not running). 

**Optional approach:**

1) Add the following line to `crontab_upps.j2`

```
#version:{{ deployment_version }}
```

2) In your new script `reload_crontab_upps.j2`, instead of checking for the signal file, you could look for version in running cronjobs. Something like

```
CRON_VER=`crontab -l | grep "#version:" | cut -d: -f2`
IRMA_VER=`echo $BASH_SOURCE | cut -d/ -f5`
IRMA_ENV=`echo $BASH_SOURCE | cut -d/ -f4`

if [ "$IRMA_ENV" == "staging" ] || [ "$CRON_VER" == "$IRMA_VER" ]; then
    exit 0
fi
```
(even if it didn't find any `version` in running `cronjobs` it should restart as one would expect it to have the `version`)

3) Rather than running the script `reload_crontab_upps` via `crontab`, it can be added to user's `.bashrc` or `sourceme_upp.sh` (as one will always reload the `env` after new version is deployed, it should do the trick).

sync.py
```diff
@@ -20,159 +20,197 @@
...
+	exp_success = ".*\$ " # Matches the end of a bash prompt 
+
+	recv = child.expect([exp_token, exp_success])
+
+	if recv == 0:
+		print('Sending SSH token') 
+		child.sendline(token)
+		print('Waiting for success')
+		child.expect(exp_success)
+		child.sendline("Logged in with password + factor, logging out.")
```
slohse: Shouldn't this be print?

johanherman: Doh! Of course! Thanks. :)

johanherman:
@slohse can I close this? :smile:

johanherman:
Tomorrow would be a suitable day... :fireworks:

slohse:
Heh. Yeah, close it :)

PR #267, opened 2018-02-16
remove path to exclude from archive

b97pla:
remove empty path from exclude_from_tarball list which caused everything to be excluded


senthil10:
@b97pla This looks like easy change to me, shall I merge it ? Or you wanna wait for @slohse

b97pla:
@senthil10 I think you can go ahead

PR #279, opened 2018-04-13
kong/lua: fix broken kong deployment

slohse:
Kong does no longer generate its start script in the bin directory of
the lua runtime when installed via luarocks. So instead we install from
source, which for the most part does the exact same thing
Meanwhile, according to the kong docs
(https://getkong.org/install/source/), it is recommended to use the lua
runtime that comes with openresty. So I also reordered the installation process a little and removed installing the standalone luajit.

### DIFF
--- a/roles/tarzan/tasks/dependencies.yml
+++ b/roles/tarzan/tasks/dependencies.yml
@@ -1,74 +1,12 @@
 ---
-# kong demands: 
-# luajit
+# kong demands:
+# openssl sources
+# openresty (comes with luajit)
 # luarocks
-# openresty
 # cassandra
 # serf
 
 # use a loop a la http://stackoverflow.com/questions/30785281/one-loop-over-multiple-ansible-tasks ? 
-
-- name: create luajit destination directory 
-  file: path={{ luajit_dest }} state=directory mode=g+rwXs
-
-- name: download luajit archive
-  get_url:
-    url: "{{ luajit_url }}"
-    dest: "{{ luajit_dest }}/{{ luajit_file }}"
-
-- name: unpack luajit to destination
-  unarchive: 
-    src: "{{ luajit_dest }}/{{ luajit_file }}" 
-    dest: "{{ luajit_dest }}"
-    copy: no 
-    creates: "{{ luajit_dest }}/{{ luajit_version }}/bin/luajit-{{ luajit_version }}"
-    mode: g+rwX,o+rX
-
-- name: set luajits install path
-  lineinfile: dest="{{ luajit_dest }}/LuaJIT-{{ luajit_version }}/Makefile"
-              regexp="^export PREFIX=./usr/local"
-              line='export PREFIX={{ luajit_dest }}/{{ luajit_version }}'
-              backup=no
-
-- name: compile luajit 
-  shell: make install
-  args: 
-    chdir: "{{ luajit_dest }}/LuaJIT-{{ luajit_version }}"
-    creates: "{{ luajit_dest }}/{{ luajit_version }}/bin/luajit-{{ luajit_version }}"
-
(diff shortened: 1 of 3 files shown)


roles/tarzan/tasks/dependencies.yml
```diff
@@ -114,6 +52,34 @@
...
+- name: unpack luarocks to destination
+  unarchive: 
+    src: "{{ luarocks_dest }}/{{ luarocks_file }}" 
+    dest: "{{ luarocks_dest }}"
+    copy: no 
+    creates: "{{ luarocks_dest }}/{{ luarocks_version }}/bin/luarocks"
+    mode: g+rwX,o+rX
+
+- name: configure luarocks
+  shell: "./configure --lua-suffix=jit --with-lua={{ openresty_dest }}/{{ openresty_version }}/luajit --with-lua-include={{ openresty_dest }}/{{ openresty_version }}/luajit/include/luajit-2.1 --prefix={{ luarocks_dest }}/{{ luarocks_version }}"
```
senthil10: Is it ok to have version *(2.1)* hard code here i.e. in `--with-lua-include={{ openresty_dest }}/{{ openresty_version }}/luajit/include/luajit-2.1`

senthil10: Maybe could have a place holder variable `luajit__include_version` *(or something)* if it is not exactly same as `luajit_version` in `defaults/main.yaml`

slohse: I was wondering about that as well, but I don't see a reliable way to figure out the luajit version bundled with openresty. I could add a variable `openresty_bundled_luajit_version` (or similar) to where the other openresty-variables are defined, so that one at least has a reminder in the vicinity to check for that when updating openresty

senthil10: ah ha.. I see your point, maybe if this path `{{ openresty_dest }}/{{ openresty_version }}/luajit/include` is going to be standard i.e. `laujit` bundled with `openresty` is always in that path, then maybe we could deduce the version like

```
- name: Grab luajit version number
  shell: "ls {{ openresty_dest }}/{{ openresty_version }}/luajit/include | grep luajit-[0-9a-z\.]*"
  register: openresty_bundled_luajit_version
```

slohse: or we just replace it with `*` and let the shell take care of it
```
shell: "./configure --lua-suffix=jit --with-lua={{ openresty_dest }}/{{ openresty_version }}/luajit --with-lua-include={{ openresty_dest }}/{{ openresty_version }}/luajit/include/luajit-* --prefix={{ luarocks_dest }}/{{ luarocks_version }}"
```

senthil10: Sure, I was just being extra paranoid 😜

roles/tarzan/tasks/dependencies.yml
```diff
@@ -114,6 +52,46 @@
...
+  find:
+    paths: "{{ openresty_dest }}/{{ openresty_version }}/luajit/include"
+    patterns: "luajit-*"
+    recurse: no
+    file_type: directory
+  register: luajit_include_folder
+
+- name: make sure the include folder is unambiguous
+  assert:
+    that: "luajit_include_folder.matched == 1"
```
slohse: Okay, like I feared, the shell-globbing didn't work. I changed it so that I use ansible's `find` module to get the luajit include folder name and a basic sanity check that makes sure it is only one folder

senthil10: Well, I think its more better this way :)

senthil10:
Nice work @slohse 👏

PR #281, opened 2018-08-30
Refactoring for next deployment

senthil10:
It looks like big PR but it isn't, I just renamed few folders and upgraded new version for some tools


host_vars/127.0.0.1/main.yml
```diff
@@ -43,7 +44,15 @@ ngi_pipeline_db_sthlm: "{{ ngi_pipeline_sthlm_path }}/db/records_db_sthlm.sql"
 piper_module_version: 1.5.1
 gatk_bundle_b37: "/sw/data/uppnex/piper_references/2016-04-07/gatk_bundle/2.8/b37/"
 
-ngi_rnaseq_dest: "{{ sw_path }}/ngi-rnaseq"
+sarek_repo: "https://github.com/SciLifeLab/sarek.git"
+sarek_version: "c63d8134bbec0cf72c534e16bd8f43c932b02193"
+sarek_tag: "latest"
```
b97pla: shouldn't we pin this to a specific tag, e.g. 2.0.0 or 2.1.0?

senthil10: Yes, initially I had only one variable `sarek_version` which I used both for `git` pull and `singularity`. I spoke with @MaxUlysse and he suggested to use latest *commit* rather than the *release*. So I split it into two variables `sarek_version` (which I use for `git`) and `sarek_tag` (which I use for `singularity` and `sarek` pipeline config).

Once there is a stable version for `sarek`, I will revert it back 👍

MaxUlysse: Sorry about that

b97pla: ok

senthil10:
Tack @b97pla :)

PR #296, opened 2018-10-31
Removed NXF_WORK definition from Irma bashrc

aanil:
Removed NXF_WORK definition from Irma bashrc.


senthil10:
IF not defined uses the directory from where `nextflow` is started and that is the behaviour we want.

PR #298, opened 2018-12-05
update ngi_pipeline version

b97pla:
This updates the ngi_pipeline to a version that supports cleaning of sarek work directory and parsing sarek results into Charon.


slohse:
:+1:

PR #300, opened 2019-01-22
Updating MultiQC and nf-core deployments

ewels:
I intended to only tweak a couple of parameters, but after reading through and figuring out how stuff works (it's really cool!) I ended up doing quite a bit of standardisation throughout a few packages:

* methylseq
* rnaseq
* neutronstar
* sarek
* MultiQC

Specifically, these things:

* I tried to move pipeline-specific variables out of the global config into their package specific configs (sarek, rnaseq)
* Standardisation of how the singularity image names are defined
* Moving everything that can be in a config into a config (with the exception of sarek)
* Adding new stuff into pipeline configs, eg. the location of the custom MultiQC config file
* Tweaking of MultiQC config params, eg. a default template and the interactive threshold

This should fix these issues that we have today:

* MultiQC reports generated by nf-core pipelines should now be automatically copied to Genomics Status
  * They didn't have the NGI-Stockholm specific config passed to them before
* The NGI template is now automatically applied, so `-t ngi` isn't needed any more
* MultiQC will generate interactive plots for projects with up to 1000 samples, instead of switching to static plots at 100 samples as before.

This is a bit more of a major change than I was intending, so it would be great if we can have a pretty thorough review / test if possible.

Note that basically all nf-core nextflow pipelines should be installed and configured in the same way. So it may be worth setting up a general nf-core rule that runs multiple times for different pipelines. This could scale better in the future. Then we can just give a list of pipelines and versions. This probably deserves a separate PR though :)


roles/neutronstar/templates/neutronstar_site.config
```diff
@@ -0,0 +1,13 @@
...
+    container = '{{ neutronstar_container_path }}/{{ neutronstar_image }}'
+    withName:multiqc {
+        container = null
+        executor = 'local'
+    }
+}
+params {
+    project = "{{ item.project_id }}"
+    multiqc_config = "{{ ngi_pipeline_conf }}/multiqc_{{ item.site }}_config.yml"
+    clusterOptions = "-A {{ ngi_pipeline_sthlm_delivery }} -p node"
```
b97pla: this should be `{{ item.project_id }}`, right?

b97pla:
Nice cleanup :+1: looks pretty straightforward to me but obviously, the changes should be carefully tested in the staging environment.

roles/neutronstar/templates/neutronstar_site.config
ewels: Correct, good spot 👍 

Looks like this error was there before (see [above](https://github.com/NationalGenomicsInfrastructure/irma-provision/pull/300/files#diff-7beecd24084a8d96a56f894e0e69a217L36)) and I just moved it over. Kind of surprised that it ever worked.. Maybe it didn't?

ewels:
@aanil / @senthil10 pointed out that my moving the Sarek config out of the default files and into its own subfolder may break stuff as one or two of those variables are used elsewhere, such as in the `ngi_pipeline` role:

https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/f8dac085e55e8266cb665f2e531e98f78a49cd82/roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2#L41-L51

I'm not confident enough to know whether this will break or not though. Or what this code is doing and whether it needs to be in that role. It seems preferable to keep things out of the global configs if possible.

ewels:
> _I'm not confident enough to know whether this will break or not though. Or what this code is doing and whether it needs to be in that role._

Specifically, the `ngi_pipeline/tasks/main.yml` file makes no mention of Sarek, so I'm not sure if those Sarek variables in `ngi_pipeline/templates/irma_ngi_config.yaml.j2` are used for anything...?

b97pla:
right, so the ngi_pipeline config file is indeed parsed and the config values are used when setting up the sarek analyses. this config handling in ngi_pipeline is not ideal and lacks a lot of flexibility, so it could do with an overhaul but right now, I'd say that we need to keep those variables in a global context

b97pla:
see  https://github.com/NationalGenomicsInfrastructure/ngi_pipeline/blob/master/ngi_pipeline/engines/sarek/models/sarek.py#L66

b97pla:
@ewels it would be great if we could get a new production environment out this week. if you revert the move of the sarek config parameters we can merge this and I can do a separate PR bumping the sarek and nextflow versions.

ewels:
Done 👍 Shout if there's anything else..

b97pla:
:+1: nice, i'll merge this now

PR #301, opened 2019-02-04
bumped sarek version, stored nextflow url and version in a parameter

b97pla:


b97pla:
great, if someone can merge (@aanil?), I can look into deploying this to staging

PR #306, opened 2019-03-28
Use nextflow snapshot

b97pla:
This PR will deploy the latest Nextflow release which should include a fix for slurm-issues. In addition, it specifies versions for MultiQC_NGI and Sisyphus.


slohse:
Looks good to me :+1:

PR #307, opened 2019-04-05
Modified Irma specific sarek config

alneberg:
Updated config file to make sure that processes using a full node is also submitted to the node queue. Remapping of these jobs by uppmax are causing errors where Nextflow loose track of the job in the slurm queue.


b97pla:
:+1:

PR #313, opened 2019-09-19
Added atacseq and ampliseq, removed nougat, bumped reports, multiqc, …

aanil:
…taca, taca_ngi_pipeline. We decided to decouple the server change from the usual deployment so that we will have no other dependencies if things go wrong.


roles/ampliseq/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+ampliseq_repo: "https://github.com/nf-core/ampliseq.git"
+ampliseq_dest: "{{ sw_path }}/ampliseq"
+ampliseq_container_url: "docker://nfcore/ampliseq:{{ ampliseq_version }}"
+ampliseq_container_path: "{{ ngi_containers }}/ampliseq"
+ampliseq_image: "ampliseq-{{ ampliseq_version }}.simg"
+ampliseq_version: "1.0.0"
```
alneberg: I shoud have told you earlier, but I just checked and there's a version 1.1.0 release available now. Let's go for that one instead?

roles/multiqc/defaults/main.yml
```diff
@@ -2,6 +2,6 @@ multiqc_repo: https://github.com/ewels/MultiQC.git
 multiqc_dest: "{{ sw_path }}/multiqc"
 multiqc_version: "v1.7"
 
-multiqc_ngi_repo: https://github.com/ewels/MultiQC_NGI.git
+multiqc_ngi_repo: https://github.com/NationalGenomicsInfrastructure/MultiQC_NGI.git
```
alneberg: 👍

PR #314, opened 2019-09-24
Methylseq config

alneberg:
Config values that we have been supplying for the methylseq at ngi-sthlm.

Might be more adjustments needed for really large projects.

### DIFF
--- a/roles/methylseq/templates/methylseq_site.config
+++ b/roles/methylseq/templates/methylseq_site.config
@@ -4,8 +4,33 @@ process {
         container = null
         executor = 'local'
     }
+    
+    /* Default memory is less than we have per cpu */
+    memory = { 15.GB * task.attempt }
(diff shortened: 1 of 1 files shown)


roles/methylseq/templates/methylseq_site.config
```diff
@@ -1,11 +1,37 @@
...
+        memory = { 125.GB * task.attempt }
+        time = { 18.h * task.attempt }
+    }
 }
 params {
     project = "{{ item.project_id }}"
     multiqc_config = "{{ ngi_pipeline_conf }}/multiqc_{{ item.site }}_config.yml"
+
+    // Irma memory is greater than default rackham
+    max_memory = '250.GB'
```
ewels: Pedantic comment: quotes not needed here.. 😀

PR #318, opened 2019-10-09
Fix syntax

aanil:
We get a syntax error without the quotes.


b97pla:
:+1: thanks for fixing

PR #321, opened 2019-10-23
Irma nextflow config

alneberg:
My take on what I think @ewels was saying today (please correct me if I misunderstood something).

This would remove the need for `-profile uppmax` and instead replace it with a general irma nextflow config file. The long term goal is to use `nf-core download` inside of the playbook to remove a lot of code.


ewels:
For bonus points, we could use this and drop `-profile uppmax` for all other nf-core pipelines, even if some don't technically need it yet (they will do soon).

alneberg:
Yes, I think it's in fact needed for a few of them, such as `atacseq` and `ampliseq`.

PR #331, opened 2019-12-05
Needed backslashes for newlines

alneberg:
Bugfix for the aliases of methylseq and rnaseq.


aanil:
Its up on genomics-status-stage

PR #332, opened 2020-01-27
New nf-core role to pull  methylseq, rnaseq, atacseq etc and other changes

aanil:
- New nf-core role to pull
  - methylseq v1.4 (Updated version as well)
  - rnaseq v1.3 (Updated version as well)
  - atacseq
  - ampliseq
- Deleted old roles for methylseq, rnaseq, atacseq and ampliseq
- Update Sarek to pull from nf-core
- Updated ngi reports
- Updated taca
- Updated neutronstar to 1.0.0
- Updated multiqc to v1.8


roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    dest: "{{ ngi_pipeline_conf }}/{{ item.0.name }}_{{ item.1.site }}.config"
+  with_nested:
+  - "{{ pipelines }}"
+  - [ { site: "sthlm", project_id: "{{ ngi_pipeline_sthlm_delivery }}" }, { site: "upps", project_id: "{{ ngi_pipeline_upps_delivery }}" }]
+
+- name: Set alias for pipeline
+  lineinfile:
+    dest: "{{ ngi_pipeline_conf }}/{{ item.1.script }}"
+    line: >
+          alias {{ item.0.name }}='nextflow run {{ sw_path }}{{ item.0.name }}/workflow/main.nf \
```
ewels: Don't need the `main.nf` bit.. (in fact, better without)
```suggestion
          alias {{ item.0.name }}='nextflow run {{ sw_path }}{{ item.0.name }}/workflow/ \
```
And just checking: `{{ sw_path }}{{ item.0.name }}` definitely shouldn't have a `/` in the middle?

roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Create directories for singularity images
+  file:
+    path: "{{ ngi_containers }}/{{ item.name }}"
+    state: directory
+    mode: g+s
+  with_items: "{{ pipelines }}"
+
+- name: Move singularity images
+  shell: "mv {{ sw_path }}{{ item.0.name }}/singularity-images/{{ item.1 }} {{ ngi_containers }}/{{ item.0.name }}/"
```
ewels: Why do we need to move the singularity images?

roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2
```diff
@@ -74,7 +74,7 @@ analysis:
             analysis_engine: ngi_pipeline.engines.qc_ngi
         RNA-seq:
             analysis_engine: ngi_pipeline.engines.rna_ngi
-            ngi_nf_path: {{ rnaseq_dest }}/main.nf
+            ngi_nf_path: {{ sw_path }}rnaseq/workflow/main.nf
```
ewels:
```suggestion
            ngi_nf_path: {{ sw_path }}rnaseq/workflow/
```

roles/nf-core/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
+nextflow_java: "/sw/comp/java/x86_64/sun_jdk1.8.0_151"
+nextflow_version_tag: "v19.04.1"
+nextflow_download_url: "https://github.com/nextflow-io/nextflow/releases/download/{{ nextflow_version_tag }}/nextflow"
+nf_core_env: "/lupus/ngi/irma3/nf-core-env"
+pipelines:
+  - name: rnaseq
+    release: 1.4.2
+    containers:
+      - nf-core-rnaseq-1.4.2.simg
```
ewels: Can we use `{{ release }}` in the container name? Could avoid accidental mismatches in the future..

ewels: Or indeed we could presumably autogenerate this filename at a later date wherever it is used:
```
nf-core-{{ pipeline_name }}-{{ pipeline_version }}.simg
```

roles/nf-core/tasks/main.yml
slohse: `sw_path` is defined in `host_vars/127.0.0.1/main.yml` with a trailing slash: `sw_path: "{{ root_path }}/sw/"`. So this should work.
Often enough we get double slashes from that - which of course is no problem technically, it just triggers me :grimacing: :smile:

roles/nf-core/defaults/main.yml
alneberg: How is this thought to work for pipelines with multiple containers? Am I right in that nf-core download currently doesn't work for those? Is there a plan for that?

roles/nf-core/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
+nextflow_java: "/sw/comp/java/x86_64/sun_jdk1.8.0_151"
+nextflow_version_tag: "v19.04.1"
```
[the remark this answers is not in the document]
aanil: Done!

roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    _raw_params: "{{ nf_core_env }}/bin/nextflow"
+  environment:
+    NXF_LAUNCHER: "/scratch"
+    NXF_HOME: "{{ nf_core_env }}/nextflow"
+    NXF_WORK: "/scratch"
+
+- name: nf-core
+  command: "{{ nf_core_env }}/bin/nf-core -v download {{ item.name }} --singularity --compress none --outdir {{ sw_path }}{{ item.name }} --release {{ item.release }}"
+  environment:
+    PATH: "{{ ansible_env.PATH }}:{{ nf_core_env }}/bin"
```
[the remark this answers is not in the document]
aanil: Done!

roles/nf-core/tasks/main.yml
aanil: Removed the `main.nf` bit. For the `sw_path`, as Stephan said, there was a trailing slash in the original file which I haven't changed.

roles/nf-core/tasks/main.yml
aanil: All the singularity images were put into a common directory in the original flow, so we just mimicked it here

roles/ngi_pipeline/templates/irma_ngi_config.yaml.j2
aanil: Done!

roles/neutronstar/tasks/main.yml
```diff
@@ -39,14 +52,19 @@
   lineinfile:
      dest: "{{ ngi_pipeline_conf }}/{{ bash_env_sthlm_script }}"
      line: >
-         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile standard,uppmax
-         -c {{ ngi_pipeline_conf }}/neutronstar_{{ item.site }}.config'
+         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile standard,uppmax \
```
remiolsen: Does this use the nf-core/configs repo? You should drop`-profile standard,uppmax` for `-profile singularity`

There are still some more options that is more suited and/or needed for irma. As a config file load with `-c`:

```
params {
  busco_folder = '/sw/apps/bioinfo/BUSCO/v2_lineage_sets'
  busco_data = 'eukaryota_odb9'
  clusterOptions = '-A ngi2016003'
  max_memory = 250.GB
  max_cpus = 16
  max_time = 240.h
}
process {
  executor = 'slurm'
}
```

Or alternatively add them as parameters to that alias with `-process.executor='slurm'`

remiolsen: I'm a bit unsure if to keep `busco_data = 'eukaryota_odb9'` - I think it's a reasonable default and does make the pipeline more foolproof.

roles/neutronstar/defaults/main.yml
```diff
@@ -4,8 +4,13 @@ neutronstar_dest: "{{ sw_path }}/neutronstar"
 # neutronstar_container_url: "docker://nf-core/neutronstar:{{ rnaseq_version }}"
```
remiolsen: Sorry, `rnaseq_version`?

aanil: Ah, that line is commented out and won't run. It was leftover from before but I can remove it if you want.

roles/neutronstar/tasks/main.yml
aanil: I've updated the parameters in this [config file](https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/94b086d5dd978683765f2a8b534c545143359b49/roles/neutronstar/templates/neutronstar_site.config). Also changed the profile to `singularity`.

roles/nf-core/tasks/main.yml
ewels: `nf-core download` pulls the singularity images and organises the files into a directory structure along with a config so that if you run the workflow from the pipeline it knows where to find the singularity image without needing a `-with-singularity <path>` command line argument.

Unless we have a specific reason for moving the singularity images, it seems that it would be better to leave them where they are..?

roles/nf-core/defaults/main.yml
ewels: Yes, you are correct. The plan is to build support for this in to `nf-core download`, it should be possible (and hopefully not too difficult). It's a fairly low priority for now though as it's not needed for many pipelines as it stands.

My suggestion was to keep the `sarek` role separate for now as it needs custom behaviour here. In the long run I would like to add this support in nf-core download and keep as much logic in there as possible to avoid having to hardcode anything here.

roles/neutronstar/tasks/main.yml
ewels: @remiolsen - you should add this config as an uppmax + pipeline specific config in the nf-core/configs repo.

See https://github.com/nf-core/configs/tree/master/conf/pipeline and https://github.com/nf-core/configs/blob/master/pipeline/ and ask @MaxUlysse for help if you need it 👍

ewels: Note that it's only `params.busco_folder` and `params.busco_data` that looks unique to me. I think everything else is already set up in the uppmax profile.

roles/nf-core/tasks/main.yml
ewels: Apologies, the config to no longer need `-with-singularity <path>` is not yet implemented. See https://github.com/nf-core/tools/issues/464

ewels:
I'm happy for this to be merged / deployed now 👍 

If we go ahead now, could we please make issues to address the following in the future though, so that they're not forgotten:

* Remove custom `neutronstar` config once @remiolsen has put the uppmax-specific stuff in to nf-core/configs
* Get rid of the singularity image `mv` stuff
    * Dependent on https://github.com/nf-core/tools/issues/464 and https://github.com/nf-core/tools/issues/513

aanil:
Created the issues #335 and #336  for it!

roles/neutronstar/tasks/main.yml
```diff
@@ -39,14 +52,19 @@
   lineinfile:
      dest: "{{ ngi_pipeline_conf }}/{{ bash_env_sthlm_script }}"
      line: >
-         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile standard,uppmax
-         -c {{ ngi_pipeline_conf }}/neutronstar_{{ item.site }}.config'
+         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile singularity \
```
ewels:
```suggestion
         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile uppmax \
```

Why are we removing `-profile uppmax`? I think if we leave this then we can also simplify the config file from @remiolsen  a lot, to this:

```nextflow
process.container = '{{ neutronstar_container_path }}/{{ neutronstar_image }}'
params {
  busco_folder = '/sw/apps/bioinfo/BUSCO/v2_lineage_sets'
  busco_data = 'eukaryota_odb9'
}
```
I think `clusterOptions` and `multiqc_config ` etc should be in the general nextflow config that we use for other pipelines.

It would be best to keep the neutronstar config file as small as possible, only covering the stuff we really need to for _just_ the neutronstar pipeline. Then everything else is consistent across all pipelines 👍

roles/neutronstar/templates/neutronstar_site.config
```diff
@@ -1,5 +1,6 @@
 process {
     container = '{{ neutronstar_container_path }}/{{ neutronstar_image }}'
+    executor = 'slurm'
```
ewels: See above comment about slimming this config file right down..

roles/nf-core/defaults/main.yml
ewels: > Can we use `{{ release }}` in the container name? Could avoid accidental mismatches in the future..

Bump 😄

roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+  with_nested:
+  - "{{ pipelines }}"
+  - [ { site: "sthlm", project_id: "{{ ngi_pipeline_sthlm_delivery }}" }, { site: "upps", project_id: "{{ ngi_pipeline_upps_delivery }}" }]
+
+- name: Set alias for pipeline
+  lineinfile:
+    dest: "{{ ngi_pipeline_conf }}/{{ item.1.script }}"
+    line: >
+          alias {{ item.0.name }}='nextflow run {{ sw_path }}{{ item.0.name }}/workflow/ \
+          --custom_config_base {{ sw_path }}{{ item.0.name }}/configs \
```
ewels: If we are using `nf-core download` then I'm pretty sure that we don't need this line.

roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+          -profile uppmax \
+          -c {{ ngi_pipeline_conf }}/nextflow_irma_{{ item.1.site }}.config \
+          -c {{ ngi_pipeline_conf }}/{{ item.0.name }}_{{ item.1.site }}.config'
```
ewels: These are the general configs I mention in relation to the neutronstar pipeline - I think they contain most of the stuff that @remiolsen had in there..

roles/nf-core/templates/site.config.j2
```diff
@@ -0,0 +1,20 @@
+process {
+  {% for i in item.0.containers %}
+  container = '{{ ngi_containers }}/{{ item.0.name }}/{{ i }}'
+  {% endfor %}
```
ewels: This won't work if there is more than one container, as we will just get `process.container` repeated multiple times. This is essentially just a variable name, so it will overwrite previous instances.

So this will only work when `item.0.containers` happens to be a list with a single element.

For more than that, the nextflow pipeline configuration needs to be more specific about where each container is used.

aanil: Removed the containers list! We'll let `nf-core download` handle multiple containers.

roles/nf-core/tasks/main.yml
aanil: Will `nf-core download` link the general site config that we create in the previous step?

alneberg: @ewels wont the pipeline start look online for configs if you skip this line?

roles/nf-core/tasks/main.yml
aanil: Updated neutronstar with the general config!

roles/neutronstar/templates/neutronstar_site.config
aanil: Done!

roles/nf-core/defaults/main.yml
aanil: Removed the container list!

roles/nf-core/tasks/main.yml
ewels: No - `nf-core download` actually edits the pipeline config files to set `params.custom_config_base` to point to the parent directory ([see code](https://github.com/nf-core/tools/blob/9e33e7f02758964a1a09436af1eb6afe0e596e4c/nf_core/download.py#L226-L243)). So as long as you leave the directory structure in tact then it should _just work_..

roles/neutronstar/tasks/main.yml
aanil: Done!

roles/nf-core/tasks/main.yml
aanil: I've removed the `custom_config_base` option.

alneberg: https://gph.is/1SHO8hI 👍

roles/neutronstar/tasks/main.yml
```diff
@@ -39,14 +52,20 @@
   lineinfile:
      dest: "{{ ngi_pipeline_conf }}/{{ bash_env_sthlm_script }}"
      line: >
-         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile standard,uppmax
-         -c {{ ngi_pipeline_conf }}/neutronstar_{{ item.site }}.config'
+         alias neutronstar='nextflow run {{ neutronstar_dest }}/main.nf -profile uppmax \
+         -c {{ ngi_pipeline_conf }}/nextflow_irma_{{ item.site }}.config \
+         -c {{ ngi_pipeline_conf }}/neutronstar_{{ item.site }}.config --supernova_container {{ supernova_container_path }}'
```
ewels: Cool 👍 

We could move `supernova_container` in to the neutronstar config file, but seeing as we will hopefully get rid of this config file before long, I think it's better how you currently have it ✅

roles/nf-core/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+nextflow_java: "/sw/comp/java/x86_64/sun_jdk1.8.0_151"
+nextflow_version_tag: "v19.10.0"
```
ewels: A new version of nextflow came out a few days ago, so you can grab that now

roles/nf-core/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+  file:
+    path: "{{ ngi_containers }}/{{ item.name }}"
+    state: directory
+    mode: g+s
+  with_items: "{{ pipelines }}"
+
+- name: Move singularity images
+  shell: "mv {{ sw_path }}{{ item.name }}/singularity-images/nf-core-{{ item.name }}-{{ item.release }}.simg {{ ngi_containers }}/{{ item.name }}/"
+  with_items:
+    - "{{ pipelines }}"
```
ewels: Didn't we say we were going to remove this, and leave the `nf-core download` dir structure?

aanil: I've created #336 for it as its dependent on some pending PRs in nf-core.

roles/nf-core/defaults/main.yml
aanil: Done!

roles/neutronstar/tasks/main.yml
aanil: 👍

aanil:
Merging with verbal approval from Phil!

PR #347, opened 2020-05-26
Fix location for ampliseq classifier

aanil:


roles/nf-core/tasks/main.yml
```diff
@@ -81,7 +81,7 @@
 - name: Download SILVA classifier data for ampliseq
   get_url:
     url: https://www.arb-silva.de/fileadmin/silva_databases/qiime/Silva_132_release.zip
-    dest: "{{ ngi_resources }}ampliseq/Silva_132_release.zip"
+    dest: "{{ ngi_rcontainers }}/ampliseq/Silva_132_release.zip"
```
ssjunnebo, line 84: Should this be `ngi_rcontainers` or `ngi_containers`?

aanil: Nice catch! Fixed it.

PR #358, opened 2020-08-20
Bio 1876 ngi pipeline adaptation to sarek 2.6

b97pla:
This PR bumps the `ngi_pipeline` commit to the first version that can launch Sarek >=2.5. It includes a couple of corresponding changes to the config file.


aanil (approved):
👍

PR #360, opened 2020-08-31
Update uppsala nextflow conf

matrulda:
This PR updates the uppsala nextflow config so that it does not contain custom parameters for the MultiQC process. 

I also added a jinja2 suffix to the template.


aanil (approved):
👍

PR #371, opened 2020-09-17
Stage sarek results with taca

matrulda:
This PR: 
 - Removes the siswrap role since it is not used anymore
 - Patch updates of two arteria webservices 
 - Adds taca configs plus resources to stage sarek analysis results, where the analysis has been started by `ngi_pipeline`
   - The sthlm version of the config is not really complete, let me know what you want to add
   - Delivery README and recalibration script were previously version controlled in the ngi_pipeline repo, but I think it makes more sense to have them here.
   - The recalibration script has been updated to use the sarek container on uppmax.


roles/sarek/templates/apply_recalibration.sh.j2
```diff
@@ -0,0 +1,59 @@
...
+
+# if running on a compute node, use the scratch area for output, otherwise write
+# directly to the output directory
+if [ -z "$SLURM_JOB_ID" ]
+then
+  SNIC_TMP="`dirname ${BAMIN}`"
+fi
+
+BAMOUT="$SNIC_TMP/`basename ${BAMIN/.bam/.recal.bam}`"
+FASTAREF="/sw/data/uppnex/ToolBox/hg38bundle/Homo_sapiens_assembly38.fasta"
```
b97pla, line 48: I have no good suggestion on how to do it, but it would be nice if the FASTAREF could be dynamic since not all projects will be mapped against 38. Perhaps having the reference as an optional third script argument with this value as the default?

roles/taca/templates/site_taca_sarek_delivery.yml.j2
```diff
@@ -0,0 +1,61 @@
...
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Preprocessing/TSV/duplicates_marked*
+            - <STAGINGPATH>/<SAMPLEID>/results/Preprocessing/TSV
+            - required: True
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Preprocessing/<SAMPLEID>/DuplicatesMarked/*
+            - <STAGINGPATH>/<SAMPLEID>/results/Preprocessing/DuplicatesMarked
+            - required: True
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Reports/*
```
b97pla, line 46:
```suggestion
            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Reports/<SAMPLEID>/*
```

roles/taca/templates/site_taca_sarek_delivery.yml.j2
```diff
@@ -0,0 +1,61 @@
...
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/pipeline_info/results_description.html
+            - <STAGINGPATH>/<SAMPLEID>/results/pipeline_info
+            - required: True
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/pipeline_info/software_versions.csv
+            - <STAGINGPATH>/<SAMPLEID>/results/pipeline_info
+            - required: True
+        -
+            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Preprocessing/TSV/duplicates_marked*
```
b97pla, line 38:
```suggestion
            - <ANALYSISPATH>/<SAMPLEID>/SarekGermlineAnalysis/results/Preprocessing/TSV/duplicates_marked*<SAMPLEID>.tsv*
```

roles/taca/templates/site_taca_sarek_delivery.yml.j2
```diff
@@ -0,0 +1,61 @@
...
+    file: "{{ ngi_pipeline_site_path }}/log/taca.log"
+
+deliver:
+    rootpath: "{{ proj_root }}/{{ ngi_pipeline_site_delivery }}/nobackup/NGI"
+    analysispath: <ROOTPATH>/ANALYSIS/<PROJECTID>
+    stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
+    operator: "{{ recipient_mail }}"
+    hash_algorithm: md5
+    files_to_deliver:
+{% if "upps" == site %}
```
b97pla, line 11: I guess this means that Sthlm has a separate command for staging reports for WGS deliveries, right?

b97pla (changes requested):
Looks good to me, I had just a few comments and suggestions for staging a few files more specifically for the current sample. Those I think would be nice to fix to avoid confusing users with non-relevant files. 

The recalibration script comment is mostly a nice to have (though a quick fix if passed as a parameter).

roles/sarek/templates/apply_recalibration.sh.j2
matrulda: Ah, very good point! I'll add it as an argument

roles/taca/templates/site_taca_sarek_delivery.yml.j2
matrulda: Great 👍

roles/taca/templates/site_taca_sarek_delivery.yml.j2
matrulda: I thought it would be nice to include report files for the "lane bams" since it is included in the MultiQC that is made for the whole project. Also, I don't see why we should remove the MultiQC reports on sample level? They are a bit unnecessary since it is basically the same info as in the project report, but IMO we should alter the sarek output as little as possible. Let me know what you think.

roles/taca/templates/site_taca_sarek_delivery.yml.j2
matrulda: Yeah, in the current state. But I'm waiting for some input from Sthlm.

roles/sarek/templates/apply_recalibration.sh.j2
matrulda: @b97pla Updated now, would be great if you could have another look.

b97pla: :+1: looks good

roles/taca/templates/site_taca_sarek_delivery.yml.j2
b97pla: ok, I realize that I'm thinking of the more general case where multiple samples have been analyzed together. In that case the reports folder will hold multiple samples. that's not how we run the analysis though and that's of course reflected in the first part of the path... my bad, all is well!

roles/taca/templates/site_taca_sarek_delivery.yml.j2
b97pla: :+1:

b97pla (approved):

matrulda:
I have made some changes to stage sthlm reports and make a separate delivery readme for each site. I talked to Jun and they need more time to figure out how they want to organize things, so we decided to merge this to not block uppsala. 

@b97pla If you think it looks good, feel free to merge.

PR #378, opened 2020-11-04
Fix matplotlib version in multiqc and bump ngi_reports

aanil:


ssjunnebo (approved):
👍

roles/multiqc/tasks/main.yml
```diff
@@ -7,6 +7,9 @@
 
 #Remember to use --upgrade flag if using commit-id
 #Using the python3 env
+-name: Temporarily fix matplotlib at v3.2.2 to get rid of warnings
+  shell: "{{ sw_path }}/anaconda/envs/{{ NGI_venv_py3_name }}/bin/pip install matplotlib==3.2.2"
```
matrulda, line 11: Maybe you can use the pip module instead of shell here? I think it's nice to use specific modules over shell commands where it is possible.

matrulda (approved):
Looks good! A minor comment that you can have look at, but I'm ready to merge if you want to.

roles/multiqc/tasks/main.yml
aanil: Done!

matrulda: Nice! 👍

PR #379, opened 2020-11-11
Update ngi-reports

aanil:
Add the refactored version of ngi-reports. It requires validation on the stage deployment.


ssjunnebo (approved):
Nice 👍

PR #380, opened 2020-11-12
Update sync to make dryrun optional and to specify a particular deployment directory

aanil:
Created a `monthly` branch to make PRs to as well. This can be merged to master before the production deployment. 

We could have a bimonthly branch for sarek/nf-core changes.

### DIFF
--- a/sync.py
+++ b/sync.py
@@ -24,14 +24,14 @@ import os
 
 # Step 0. Settings to fetch from user or set globally for the sync.
 parser = argparse.ArgumentParser()
-parser.add_argument('environment', choices=['production', 'staging'], help='which environment to sync over')
-parser.add_argument('-d', '--destination', help='the non-standard destination path on the remote host to sync to')
-parser.add_argument('deployment', help='which deployment to sync over')
+parser.add_argument('-e','--environment', choices=['production', 'staging'], required=True, help='which environment to sync over')
+parser.add_argument('-d','--deployment', help='which deployment to sync over')
+parser.add_argument('--destination', help='the non-standard destination path on the remote host to sync to')
 parser.add_argument('--dryrun', action='store_true', dest='dryrun', default=False, help='do a dry run first before the actual sync')
 args = parser.parse_args()
 
-ngi_root = '/lupus/ngi/.' #break url for relative path
-src_root_path = os.path.join(ngi_root, args.environment)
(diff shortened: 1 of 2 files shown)


sync.py
```diff
@@ -23,30 +24,35 @@
 
 # Step 0. Settings to fetch from user or set globally for the sync.
 parser = argparse.ArgumentParser()
-parser.add_argument("environment", choices=["production", "staging"], help="which environment to sync over")
-parser.add_argument("-d", "--destination", help="the non-standard destination path on the remote host to sync to")
+parser.add_argument('environment', choices=['production', 'staging'], help='which environment to sync over')
+parser.add_argument('-d', '--destination', help='the non-standard destination path on the remote host to sync to')
+parser.add_argument('deployment', help='which deployment to sync over')
```
matrulda, line 29: Personally I try avoid having two positional arguments. Mainly because it requires me to remember in which order I should supply them. 😅 Suggestion: Add argument flags for both `environment` and `deployment` . If you decide to do this, remember to update the sync commands in the README.

sync.py
```diff
@@ -107,65 +113,67 @@ def yes_or_no(question):
...
                                                                                                src_containers_path,
                                                                                                user,
                                                                                                host,
                                                                                                dest)
 # TODO: Do this cleaner?
-dry_cmd = "/bin/rsync --dry-run -avzP --omit-dir-times {0} {1} {2} {3}@{4}:{5}".format(excludes, src_root_path, src_containers_path, user, host, dest)
+if args.dryrun:
+	dry_cmd = '/bin/rsync --dry-run -avzP --relative --omit-dir-times {0} {1} {2} {3}@{4}:{5}'.format(excludes, src_root_path, src_containers_path, user, host, dest)
 
-# First doing a dry-run to confirm sync.
```
matrulda, line 144: Maybe we can remove this comment since it isn't always the first thing we do.

matrulda (approved):
Looks good to me! Left some minor comments, let me know if you want to address them.

sync.py
aanil: Done! I wasn't quite sure of how the deployment directory looks like in arteria's case, but that option is not required to sync.

sync.py
aanil: Done!

ssjunnebo (approved):
Looks good! 👍

PR #384, opened 2020-12-01
Make playbooks and roles compatible with Ansible 2.9

slohse:
This PR makes our roles and playbooks compatible with Ansible 2.9.
While I was at it, I also made some style changes and replaced all invocations of `pip` via shell with the ansible `pip` module.

On irma I installed ansible-2.9 besides our current (2.0) installation and symlinked `/lupus/ngi/irma3/ansible-env` to the ansible-2.0 installation. Once we merge this PR, we need to change that symlink to point to the 2.9 installation instead.

### DIFF
--- a/roles/ngi_reports/tasks/main.yml
+++ b/roles/ngi_reports/tasks/main.yml
@@ -9,10 +9,9 @@
        recursive=yes
 
 - name: Install ngi_reports
-  shell: "{{ sw_path }}/anaconda/envs/{{ NGI_venv_py3_name }}/bin/pip install -e ."
-  args:
-    chdir: "{{ ngi_reports_dest }}"
-
+  pip:
+    name: "file:{{ ngi_reports_dest }}"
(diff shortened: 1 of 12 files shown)


roles/ngi_reports/tasks/main.yml
```diff
@@ -9,10 +9,9 @@
        recursive=yes
 
 - name: Install ngi_reports
-  shell: "{{ sw_path }}/anaconda/envs/{{ NGI_venv_py3_name }}/bin/pip install -e ."
-  args:
-    chdir: "{{ ngi_reports_dest }}"
-
+  pip:
```
aanil, line 12: I'm not sure if this gets installed the right way. I tried running it but got an error. Maybe `executable` works better instead of `virtualenv`?

slohse: Good catch!
The fix is embarrassingly simple and took entirely too long to find :grimacing:

aanil (approved):
Had a successful test deployment!

PR #386, opened 2021-01-18
Update to rnaseq3 with new temp role, bump nextflow to v20.10.0

aanil:
- Created a new temporary role to install rnaseq v3 until `nf-core download` is updated to handle biocontainers. 
- rnaseqv3 also requires nextflow v20.10.0.

### DIFF
--- a/roles/nextflow/defaults/main.yml
+++ b/roles/nextflow/defaults/main.yml
@@ -1,3 +1,3 @@
 nextflow_java: "/sw/comp/java/x86_64/sun_jdk1.8.0_151"
-nextflow_version_tag: "v20.01.0"
+nextflow_version_tag: "v20.10.0"
 nextflow_download_url: "https://github.com/nextflow-io/nextflow/releases/download/{{ nextflow_version_tag }}/nextflow"
--- a/roles/nf-core-biocontainer-temp/tasks/main.yml
+++ b/roles/nf-core-biocontainer-temp/tasks/main.yml
@@ -16,12 +16,14 @@
 
 - name: Create temp directories for download
   file:
-    path: "{{ ngi_containers }}/download"
+    path: "{{ item }}"
     state: directory
+  with_items:
+    - "{{ ngi_containers }}/download"
+    - "{{ ngi_containers }}/download/biocontainers"
 
 #Download these for rnaseq 3 containers until nf-core dowload is updated
 #wget -L https://dl.dropboxusercontent.com/s/5venz3l2wihaqdp/nf-core-rnaseq_v3.0.tar.gz
-#wget -L https://dl.dropboxusercontent.com/s/0w2e72pz2iuj37e/nf-core-rnaseq_v3.0.tar.gz.md5
 
 - name: Download containers
   get_url:
@@ -32,14 +34,21 @@
 - name: Unarchive downloaded file
   unarchive:
     src: "{{ ngi_containers }}/download/nf-core-rnaseq_v3.0.tar.gz"
-    dest: "{{ ngi_containers }}/biocontainers"
+    dest: "{{ ngi_containers }}/download/biocontainers"
     remote_src: yes
+
+- name: Move unarchived files to biocontainers
+  synchronize:
+    src: "{{ ngi_containers }}/download/biocontainers/nf-core-rnaseq_v3.0/"
+    dest: "{{ ngi_containers }}/biocontainers/"
+
 
 - name: Clean download
   file:
     path: "{{ item }}"
     state: absent
   with_items:
+    - "{{ ngi_containers }}/download/biocontainers/"
     - "{{ ngi_containers }}/download/nf-core-rnaseq_v3.0.tar.gz"
     - "{{ ngi_containers }}/download"
 
--- a/roles/nf-core-biocontainer-temp/tasks/main.yml
+++ b/roles/nf-core-biocontainer-temp/tasks/main.yml
@@ -16,11 +16,8 @@
 
 - name: Create temp directories for download
   file:
-    path: "{{ item }}"
+    path: "{{ ngi_containers }}/download/biocontainers"
     state: directory
-  with_items:
-    - "{{ ngi_containers }}/download"
-    - "{{ ngi_containers }}/download/biocontainers"
 
 #Download these for rnaseq 3 containers until nf-core dowload is updated
 #wget -L https://dl.dropboxusercontent.com/s/5venz3l2wihaqdp/nf-core-rnaseq_v3.0.tar.gz
@@ -41,7 +38,6 @@
   synchronize:
     src: "{{ ngi_containers }}/download/biocontainers/nf-core-rnaseq_v3.0/"
     dest: "{{ ngi_containers }}/biocontainers/"
-
 
 - name: Clean download
   file:
--- a/roles/nf-core-biocontainer-temp/tasks/main.yml
+++ b/roles/nf-core-biocontainer-temp/tasks/main.yml
@@ -1,0 +1,72 @@
+---
+- name: install nf-core/rnaseq
+  command: "{{ nf_core_env }}/bin/nf-core -v download {{ item.name }}
+   --compress none --outdir {{ sw_path }}{{ item.name }} --release {{ item.release }}"
+  environment:
+    PATH: "{{ nf_core_env }}/bin:{{ ansible_env.PATH }}"
+  args:
+    chdir: "{{ sw_path }}"
+  with_items: "{{ pipelines }}"
+
+- name: Create directory for rnaseq biocontainers
+  file:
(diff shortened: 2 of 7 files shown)


roles/nf-core-biocontainer-temp/templates/site.config.j2
```diff
@@ -0,0 +1,4 @@
+{% if item.0.name=='rnaseq' %}
```
matrulda, line 1: Could you add site to this if-statement as so that seq_center can be set to Uppsala for us?

roles/ngi_pipeline/templates/sourceme_common.sh.j2
```diff
@@ -27,3 +27,5 @@ export PS1="$IRMA_PROMPT [\u@\h \w]$ "
 
 # Default SNIC_TMP as /scratch
 export SNIC_TMP="/scratch"
+
+export NXF_SINGULARITY_CACHEDIR="{{ ngi_containers }}/biocontainers/nf-core-rnaseq_v3.0/"
```
matrulda, line 31: I wonder if this could interfere with other pipelines? Or will the `container ` parameter used for other pipelines take precedence over this?

roles/nf-core-biocontainer-temp/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    mode: 02775 #ug=rwx, o=rx, g+s
+
+- name: Create temp directories for download
+  file:
+    path: "{{ ngi_containers }}/download"
+    state: directory
+
+#Download these for rnaseq 3 containers until nf-core dowload is updated
+#wget -L https://dl.dropboxusercontent.com/s/5venz3l2wihaqdp/nf-core-rnaseq_v3.0.tar.gz
+#wget -L https://dl.dropboxusercontent.com/s/0w2e72pz2iuj37e/nf-core-rnaseq_v3.0.tar.gz.md5
```
matrulda, line 24: We don't download the checksum, so maybe skip the comment about it.

roles/nf-core-biocontainer-temp/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+- name: Download containers
+  get_url:
+    url: https://dl.dropboxusercontent.com/s/5venz3l2wihaqdp/nf-core-rnaseq_v3.0.tar.gz
+    dest: "{{ ngi_containers }}/download/nf-core-rnaseq_v3.0.tar.gz"
+    remote_src: yes
+
+- name: Unarchive downloaded file
+  unarchive:
+    src: "{{ ngi_containers }}/download/nf-core-rnaseq_v3.0.tar.gz"
+    dest: "{{ ngi_containers }}/biocontainers"
```
matrulda, line 35: I think it would be nice if the rnaseq containers were put directly in `biocontainers` and not in in subfolder `nf-core-rnaseq_v3.0`. This is under the assumption that other pipeline will share the same containers in the future. But I don't know if it's do-able directly in the unarchive module and hence might require some extra steps. Maybe not worth it. Let me know what you think.

matrulda (commented):
Great work! Had a few comments you can have a look at.

roles/nf-core-biocontainer-temp/templates/site.config.j2
aanil: Done!

roles/ngi_pipeline/templates/sourceme_common.sh.j2
aanil: I think the older pipelines don't refer to this parameter yet. I checked with methylseq and it seems to run with `NXF_SINGULARITY_CACHEDIR` set. I'm guessing the other pipelines will be converted to use this parameter in the future. @ewels probably has a better idea about this?

roles/nf-core-biocontainer-temp/tasks/main.yml
aanil: Done!

roles/nf-core-biocontainer-temp/tasks/main.yml
aanil: Done! I though as the other pipelines would be downloaded through `nf-core download` most probably, it could be better to keep this separate. But you're right, it would be better to start using it the right way straightaway.

roles/ngi_pipeline/templates/sourceme_common.sh.j2
ewels: This is a core Nextflow option and so will affect all Nextflow pipelines: https://www.nextflow.io/docs/latest/config.html#environment-variables

It sets where the Singularity images are saved when they are downloaded from the internet. On a simpler system with an internet connection, you would just run any nf-core pipeline with `-profile singularity` and the images would automatically be downloaded to this folder.

Because we run offline, we can't download images, so we've never needed to set this. All of our existing pipelines refer to a _specific file path_ for their singularity containers - Nextflow doesn't download them and so no cachedir is used.

In the new `rnaseq` pipeline, we now have loads of containers instead of just one. Rather than having to manually set the file path for every container in the pipeline (would get boring and error prone quickly, especially with more pipelines), we let Nextflow do its thing with trying to download the images. However, we circumvent the problem of no internet connection by downloading the images to the cache directory ahead of time. So when Nextflow realises that it needs the image it looks there first, finds the image, and doesn't try to download it.

simples

ewels: TL;DR; - yes it will technically affect every Nextflow pipeline we have, but because they should all be specifying direct paths for the images it should not change the way that they run or present any problems.

I think.

matrulda: Cool cool! I guess it's pretty easy to test this by running another pipeline and see how it goes.

roles/nf-core-biocontainer-temp/tasks/main.yml
```diff
@@ -16,12 +16,14 @@
 
 - name: Create temp directories for download
   file:
-    path: "{{ ngi_containers }}/download"
+    path: "{{ item }}"
     state: directory
+  with_items:
+    - "{{ ngi_containers }}/download"
+    - "{{ ngi_containers }}/download/biocontainers"
```
matrulda, line 23: According to the docs only this path is needed. `download` should be created if it does not exist. But I'm looking at new docs, so maybe it's not supported in the ansible version we are currently using.

roles/nf-core-biocontainer-temp/tasks/main.yml
```diff
@@ -32,14 +34,21 @@
...
-    dest: "{{ ngi_containers }}/biocontainers"
+    dest: "{{ ngi_containers }}/download/biocontainers"
     remote_src: yes
 
+- name: Move unarchived files to biocontainers
+  synchronize:
+    src: "{{ ngi_containers }}/download/biocontainers/nf-core-rnaseq_v3.0/"
+    dest: "{{ ngi_containers }}/biocontainers/"
+
+
```
matrulda, line 45: Extra space here.

matrulda (commented):
Nice! Just two minor comments you can have a look at, then I'm happy.

roles/nf-core-biocontainer-temp/tasks/main.yml
aanil: Removed the space!

roles/nf-core-biocontainer-temp/tasks/main.yml
aanil: Updated it! We'll have the ansible v2.9 changes on the same branch, so we can go with what it says in the latest docs.

roles/nextflow/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 nextflow_java: "/sw/comp/java/x86_64/sun_jdk1.8.0_151"
-nextflow_version_tag: "v20.01.0"
+nextflow_version_tag: "v20.10.0"
```
ewels, line 2: I think that rnaseq v3 needs the latest edge release. However, that was to fix a problem with singularity image download I think so hopefully should be fine with this version. Bear in mind for testing though 👍🏻

ewels (approved):
Looks good! 🚀

roles/nextflow/defaults/main.yml
aanil: We'll have a validation with the stage deployment first, so if `v20.10.0` doesn't work out, we can update to the latest edge release and test again.

ewels: Sounds good. I think that the next stable release should be out pretty soon - in a week or two maybe.

ewels:
FYI, made a start on the `nf-core download` upgrade: https://github.com/nf-core/tools/pull/832

Hopefully will be merged / released before too long. But let's not wait for it.

PR #391, opened 2021-02-02
Bimonthly: fix arteria services

matrulda:
Fixes a problem where virtual envs for `arteria-checksum` and `archive-upload` would be created with python 3 rather than 2.

Also: Fixed "v" missing from nextflow version and updated synchronize parameters to avoid issues with permissions.


aanil (approved):
Looks good 👍

PR #394, opened 2021-02-05
Update ngi_pipeline version to increase default qc walltime

aanil:
A tiny change to include in the monthly release.


matrulda (approved):
👍

PR #397, opened 2021-03-08
Process name change on rnaseqv3.0

aanil:
Not sure how to make it case insensitive, so this seemed like the next best thing.


pericsson (approved):
Looks good!

PR #412, opened 2021-04-19
Updated sthlm sarek delivary yml to match delivery readme

pericsson:
Updating the Sarek delivery yaml for Sthlm to match the currently used [delivery readme](https://github.com/SciLifeLab/taca-ngi-pipeline/blob/master/delivery_readmes/DELIVERY.README.Sarek.txt)

### DIFF
--- a/roles/taca/templates/site_taca_sarek_delivery.yml.j2
+++ b/roles/taca/templates/site_taca_sarek_delivery.yml.j2
@@ -7,8 +7,8 @@ deliver:
     stagingpath: <ROOTPATH>/DELIVERY/<PROJECTID>
     operator: "{{ recipient_mail }}"
     hash_algorithm: md5
-    files_to_deliver:
 {% if "upps" == site %}
+    files_to_deliver:
         -
             - <ANALYSISPATH>/seqreports/*
             - <STAGINGPATH>/00-Reports/SequenceQC/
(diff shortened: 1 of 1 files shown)


roles/taca/templates/site_taca_sarek_delivery.yml.j2
```diff
@@ -68,3 +57,64 @@ deliver:
...
+            - <STAGINGPATH>/00-Reports
+        -
+            - <ANALYSISPATH>/sarek_multiqc_report/*multiqc_report.html
+            - <STAGINGPATH>/00-Reports
+        -
+            - {{ ngi_resources }}/TACA/{{ site }}/DELIVERY.README.SAREK.txt
+            - <STAGINGPATH>
+            - required: True
+    save_meta_info: True
+    xmlgen_ignore_lib_prep: False
```
aanil, line 119: `xmlgen_ignore_lib_prep` isnt needed anymore and can be removed.

pericsson: Thanks, I'll remove it.

roles/taca/templates/site_taca_sarek_delivery.yml.j2
```diff
@@ -68,3 +57,63 @@ deliver:
...
+            - {{ ngi_site_softlinks }}/ACKNOWLEDGEMENTS.txt
+            - <STAGINGPATH>
+        -
+            - <ANALYSISPATH>/reports/*
+            - <STAGINGPATH>/00-Reports
+        -
+            - <ANALYSISPATH>/*multiqc_report.html
+            - <STAGINGPATH>/00-Reports
+        -
+            - <ANALYSISPATH>/sarek_multiqc_report/*multiqc_report.html
```
jun-wan, line 112: This multiqc report is manually generated for all samples (not from the pipeline). I am not sure if we should keep it. Maybe @parlundin could make a decision?

pericsson: I agree with you that we might not need two multiqc reports.

pericsson:
`no_digest: True` was added to Sarek delivery yaml in order to not calculate the md5 sum for this file.

pericsson:
The soft link that became broken during delivery was fixed by moving the readme from `ngi_resources` to `ngi_site_softlinks`. This also fixed the md5 calculation since it is now writing it to a directory that the funk user has permission to write to.

aanil (approved):
👍

parlundin (approved):

PR #414, opened 2021-05-04
Update Sarek version to 2.7

aanil:
Pull from master to bimonthly and update to Sarek v2.7.


slohse (approved):
Oh wow, for a moment I was confused by the huge number of commits in here, but it's mostly due to the accidental merge and revert.

Looks good to me!

(I wonder if there was a more elegant way to bring the `bimonthly` branch up to the current state of `master`. If it were a personal branch I'd rebase, but that is obviously not what one would want to do with a public branch.)

pericsson (commented):
Looks good!

pericsson (approved):

aanil:
Merging with two approvals!

PR #420, opened 2021-06-08
Cleanup unused links ngi pipeline

pericsson:
The [ngi_pipeline repo](https://github.com/NationalGenomicsInfrastructure/ngi_pipeline/tree/master/scripts) does no longer contain the scripts `applyRecalibration.sh` and `bam2fastq.sh`, therefore we can remove the softlink creation from `create_ngi_pipeline_dirs.sh.j2`.


slohse (approved):

aanil (approved):
👍 Looks good!

roles/taca/templates/site_taca_delivery.yml.j2
```diff
@@ -94,10 +88,6 @@ deliver:
             - <ANALYSISPATH>/piper_ngi/05_processed_alignments/<SAMPLEID>.pre_recal.table
             - <STAGINGPATH>/<SAMPLEID>/03-BAM
             - required: True
-        -
-            - {{ ngi_site_softlinks }}/applyRecalibration.sh
-            - <STAGINGPATH>/<SAMPLEID>/03-BAM
-            - required: True
```
b97pla, line 100: Actually, in NGI-U we deliver BAM files before recalibration and therefore we need this script. The script is now distributed by the sarek role: https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/master/roles/sarek/templates/apply_recalibration.sh.j2

b97pla: Then again, I think we as well had problems with permissions when accessing the configs directly.. Perhaps I'll just fix this so that the script is linked to the softlink folder similar to the READMEs..

b97pla (changes requested):
Great with a cleanup, although I believe we need the recalibration script. If you're ok with it, I can see if I can PR your repo with a fix.

roles/taca/templates/site_taca_delivery.yml.j2
b97pla: Actually, I just realized that this was just for the old piper stuff.. Nevermind, then it's all good!

b97pla (approved):
:+1:
Looks good, ignore my previous comments, I misunderstood which config you modified, sorry!

pericsson:
Thanks for double checking @b97pla :+1:

PR #423, opened 2021-08-03
Update nfcore tools to v2.1

pericsson:
Updating nf-core tools to version 2.1.

# Updates
- Removed the `nf-core-biocontainers-temp` role by moving the `rnaseq` pipeline download into the `nf-core` role.
- Downloading singularity images using [`nf-core tools`](https://github.com/NationalGenomicsInfrastructure/irma-provision/pull/423/files#diff-6bae7cc64964a6dc20b3da0b8de7bbca3acfd78ae2790c220870075dff0702d6R44-R57) instead of `singularity pull`.
  - Images that cannot be found in the `NXF_SINGULARITY_CACHEDIR` are downloaded.
- Used the [`singularity.cacheDir`](https://github.com/NationalGenomicsInfrastructure/irma-provision/pull/423/files#diff-d5edb2311ea6237af8ae7b95dec52ca91a06be30639750241534605a5eaeccdfR27-R31) instead of `process.container` in the pipeline specific configs since it can search in a directory for all associated images.
- The singularity images of the pipelines using dsl2 and biocontainers (`ampliseq`, `viralrecon`, `rnaseq`) were all added to the same directory [`biocontainers`](https://github.com/NationalGenomicsInfrastructure/irma-provision/pull/423/files#diff-5797f8e18253b82a3c9adccee25f43f3239fe2ddd6dc2d027475d122a1b654aeR16-R24) to avoid downloading duplicates.
- Changed the directory name for the [nf-core virtual environment](https://github.com/NationalGenomicsInfrastructure/irma-provision/compare/bimonthly...pericsson:update_nfcore_tools#diff-5797f8e18253b82a3c9adccee25f43f3239fe2ddd6dc2d027475d122a1b654aeR4) and set the `nf-core tools` version to 2.1.
- Changed `nf-core tools` `download` flag `release` to [`revision`](https://github.com/NationalGenomicsInfrastructure/irma-provision/pull/423/files#diff-21e3144772f4a4251d47b5fbfd2f792f76bf4dffe8fbf079cefeac5f7e2de323R15) since the flag changed name.
- Updated nf-core tools, Nextflow and pipeline versions.
- Merged master branch into bimonthly.

# Version updates
- [methylseq](https://github.com/nf-core/methylseq) (`v1.6` (bimonthly) -> **`v1.6.1`(master +thisPR)**)
- [ampliseq](https://github.com/nf-core/ampliseq) (`v1.1.2` (bimonthly) -> `v1.2.0` (master) -> **`v2.0.0`(thisPR)**)
- [atacseq](https://github.com/nf-core/atacseq) (`v1.1.0` (bimonthly) -> `v1.2.1` (master) -> **`v1.2.1` (thisPR)**)
- [viralrecon](https://github.com/nf-core/viralrecon) (`v1.1.0` (bimonthly+master) -> **`v2.2` (thisPR)**)
- [rnaseq](https://github.com/nf-core/rnaseq) (`3.0` (bimonthly+master) -> **`3.3` (thisPR)**)
- [nf-core tools](https://github.com/nf-core/tools) (`1.7` (bimonthly+master) -> **`2.1` (thisPR)**)
- [nextflow](https://github.com/nextflow-io/nextflow) (`v20.11.0-edge` (master) -> `v21.04.1` (bimonthly) -> **`v21.04.3` (thisPR)**)
- [archive_upload_version](https://github.com/Molmed/snpseq-archive-upload) (`v1.0.2` (bimonthly) -> **`v1.0.4` (master+thisPR)**)
- [multiqc](https://github.com/ewels/MultiQC) (`v1.10` (bimonthly) -> **`v1.10.1` (master+thisPR)**)


roles/multiqc/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 multiqc_repo: https://github.com/ewels/MultiQC.git
 multiqc_dest: "{{ sw_path }}/multiqc"
-multiqc_version: "v1.10"
+multiqc_version: "v1.10.1"
```
aanil, line 3: We can update Multiqc on monthly. Also the latest version is 1.11 right now.

roles/nf-core/defaults/main.yml
```diff
@@ -1,17 +1,24 @@
...
-  - name: viralrecon
-    release: 1.1.0
+    release: 1.2.1
   - name: rnafusion
     release: 1.2.0
   - name: nanoseq
     release: 1.1.0
+  - name: ampliseq
+    release: 2.0.0
+    modified_singularity_cachedir: "{{ biocontainers_dirname }}"
```
aanil, line 18: Instead of repeating `biocontainers_dirname`, we could maybe have a flag marking whether the pipeline isDSL2 or not, since thats what it seems to be used to check in the site config, if I'm not wrong?

roles/taca/defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 taca_ngi_repo: https://github.com/SciLifeLab/taca-ngi-pipeline.git
 taca_ngi_dest: "{{ sw_path }}/taca-ngi-pipeline"
-taca_ngi_version: 1d7ca548099ed3b08dd33fff59f6b8652b323707
+taca_ngi_version: 79d6f1b38dd5808b0ab4a3e7a732c989e5cc7ec1
```
aanil, line 4: This would go in monthly too.

roles/multiqc/defaults/main.yml
pericsson: thanks, updated it to the latest now

roles/nf-core/defaults/main.yml
pericsson: I actually thought it would be more flexible if you can set the singularity cachedir to the directory name that you want, e.g. if you want to have a separate folder for a few pipelines but another for other ones? So what the config does to achieve this is that it sets this modified directory name (if you do not want to use the pipeline name) as the singularity cachedir, so that it can be variable. What do you say? Do you think there are any advantage of making it variable or should we substitute it to a boolean flag?

roles/taca/defaults/main.yml
pericsson: This version is available in master and monthly, it is only the bimonthly branch that had an older version.

aanil: Ah I misunderstood. This is fine!

roles/nf-core/defaults/main.yml
pericsson: I'll change it to a flag

roles/multiqc/defaults/main.yml
pericsson: reverted the multiqc update so that we can include it in the monthly PR instead.

roles/nf-core/defaults/main.yml
pericsson: hmm, started changing the solution into a boolean flag but realised that the way I solved it in the beginning made this part pretty tidy: https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/0fadf3ecbf4a9a483615741149cd567b48b5561d/roles/nf-core/tasks/main.yml#L53 

compared to changing the flag `modified_singularity_cachedir` to a boolean and setting the same line to
```
container_dir: "{{ biocontainers_dirname if item.modified_singularity_cachedir is defined and item.modified_singularity_cachedir == True else item.name}}"
```
However I'm still open to a nice solution if you have any ideas :)

roles/nf-core/templates/site.config.j2
```diff
@@ -1,5 +1,4 @@
 process {
```
aanil, line 1: This should be in an if with (methylseq or ampliseq) as it will give you an empty process for other pipelines which will cause an error when you run them.

roles/nf-core/defaults/main.yml
aanil: It should be fine to have it as it is then.

aanil (approved):
Looks good!

roles/multiqc/defaults/main.yml
aanil: 👍

slohse (approved):

PR #425, opened 2021-09-10
Update multiqc to 1.11 and bump TACA version

aanil:
Updates

- Multiqc to v1.11
- Update TACA version

@ssjunnebo Should we include the recent updates to taca-ngi-pipeline too?


ssjunnebo:
> @ssjunnebo Should we include the recent updates to taca-ngi-pipeline too?

Hm, not sure. It would need validation I think, so maybe it should go in bimonthly instead?

ssjunnebo (approved):
👍

PR #426, opened 2021-09-15
Move to python3

ssjunnebo:
* Switch NGI env to python3
* Remove StatusDB module which is no longer used
* Bump version of taca-ngi-pipeline to support python3

Beware that this is the first time I've worked with Ansible so I probably missed something 😅


roles/ngi_pipeline/tasks/main.yml
```diff
@@ -134,7 +134,7 @@
 
 - name: Add py3 anaconda env path to sthlm sourceme script
   lineinfile: dest={{ ngi_pipeline_conf }}/{{ bash_env_sthlm_script }}
-              line='export PATH={{ anaconda_path }}/envs/{{ NGI_venv_py3_name }}/bin:$PATH'
+              line='export PATH={{ anaconda_path }}/envs/{{ NGI_venv_name }}/bin:$PATH'
```
aanil, line 137: We don't really need to keep this if the py3 env is the main env, but no harm in leaving it in, I guess.

ssjunnebo: Ah, no you're right, let's remove it 👍

aanil (approved):
👍 Looks ok to me. Lets have a deployment and see how it goes!

PR #428, opened 2021-09-16
Remove Sthlm Sarek WES config for now

aanil:
We will just remove the Sarek WES config for Sthlm for now. We can add it back in the future if we need it.


b97pla (approved):
:+1:

ssjunnebo (approved):

PR #433, opened 2021-11-03
Py3 and nf fixes

ssjunnebo:
* Small encoding fixes in TACA and ngi_pipeline
* Nextflow version for offline usage to avoid issues with plugins


aanil (approved):
👍

PR #435, opened 2021-11-10
Set up required plugins dir for nextflow

ssjunnebo:
Nextflow tries to create a plugins directory if it doesn't exist, but only users in the ngi-sw group have write access.


aanil (approved):
👍

PR #436, opened 2022-02-07
Python3 and nextflow

ssjunnebo:
* Nextflow 21.04.3-all
* ampliseq 2.0.0
* viralrecon 2.2
* nf-core tools 2.1
* rnaseq 3.3
* python2 -> python3


ssjunnebo:
The click version is still fixed to 7.0, so I'll try removing that. Maybe that will solve your issue @slohse 🤞

slohse:
Hm, #439 didn't actually fix my issue (I cherry-picked it into my local copy of this branch to verify). But I think it's a result of the unusual locale (LANG and LC_CTYPE are `en_DE.UTF-8`) I have on my desktop machine, that SSH apparently carries over to irma. If I do `export LANG=en_US.UTF-8; unset LC_CTYPE;` in my shell before I run ansible, it works fine. It's just weird that the click branch worked without changing my Locale, whereas this branch doesn't.

slohse (approved):
Like I wrote in my comment, I suddenly got issues apparently caused by my Locale, but I don't see this as a deal breaker for this MR. We could add a note in the README, though.

aanil (approved):
👍  Looks good to me.

